PR #3, opened 2019-12-18
[POC] Use spark for hive queries

ottomata:


nettrom:
Ran some tests with this, and from what I can tell it's a good starting point (see [this phab comment](https://phabricator.wikimedia.org/T240890#5759663) for more information). Merging so we can fix a couple of outstanding issues and test more widely within the team.

PR #5, opened 2020-01-03
Check update

bearloga:
While I was reading Marshall's comment in [T240890#5772652](https://phabricator.wikimedia.org/T240890#5772652) it occurred to me that wmfdata may sometimes be updated in a major way (e.g. to support a new environment) and that the end user may not remember to occasionally upgrade the local install, especially if they don't see any update announcements.

To that end, I propose this logic for comparing the local version of the package with the remote version on `import wmfdata`. If the remote version is newer, the user sees a message like:

```
You are using wmfdata 0.1.0. A newer version is available.
Update to 0.1.1 via: pip install --upgrade git+https://github.com/neilpquinn/wmfdata/wmfdata.git
To see what changed refer to https://github.com/neilpquinn/wmfdata/CHANGELOG.md

You can find the source for `wmfdata` at https://github.com/neilpquinn/wmfdata
```

**Topic for discussion**: ability to suppress this check by defining & exporting an environmental variable in ~/.bashrc or ~/.bash_profile. Then `__init__.py` checks if the variable is defined and only performs the check if it's not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea/
 .vscode/
+wmfdata/__pycache__/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,2 @@
-# wmfdata 0.1.1
-
-- Added `check_remote_version()` in `utils` module.
-- Added a check of remote version on import and a notification that an update is available.
-
-# wmfdata 0.1.0
-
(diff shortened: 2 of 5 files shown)


nshahquinn:
Looks great! I just tweaked things a bit based on my plan to create a "release" branch rather than having people update to whatever the latest master is. That way we can do proper semantic versioning and release backwards-compatible code if necessary.

PR #6, opened 2020-01-12
Read from MariaDB charset using UTF-8

nshahquinn:
We currently tell the python-mysql-connector package to read from our MariaDB databases using the binary character set. However, for some reason this stopped working in version 8.0.18 of the connector (https://phabricator.wikimedia.org/T242448).

It turns out that although our MariaDB databases are set to use the binary character set, binary isn't really a character set at all since at the root _all_ characters sets use bytes. It's just a way of telling the database system to treat the bytes just as bytes without attempting to interpret them using a specific character set; for some reason, MediaWiki wants its databases to do to that even though the bytes actually represent text encoded with the UTF-8 character set.

So, we can more usefully tell the connector to read from the databases using UTF-8. This fixes the problem that arose with 8.0.18 and additionally causes the connector to handle the decoding of bytes to characters, allowing us to remove the code where we handled that decoding ourselves.

### DIFF
--- a/wmfdata/mariadb.py
+++ b/wmfdata/mariadb.py
@@ -21,22 +21,6 @@ def ensure_list(str_or_list):
         return [str_or_list]
     else:
         return str_or_list
-    
-
-# Strings are stored in MariaDB as BINARY rather than CHAR/VARCHAR, so they need to be converted.
-# To-do: move these decode functions to the utils module
-def try_decode(cell):
-    try:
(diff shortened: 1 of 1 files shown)


bearloga (approved):
Looks good, thank you! I tried `SELECT * FROM revision LIMIT 10;` with `date_col = "rev_timestamp"` as a test and everything worked perfectly. LGTM!

PR #7, opened 2020-01-15
Spark improvements

nshahquinn:
This patch:
* stops Spark sessions after an hour of inactivity
* improves the pre-query Kerberos credentials check
* restores the ability to pass multiple commands to hive.run

### DIFF
--- a/wmfdata/hive.py
+++ b/wmfdata/hive.py
@@ -14,37 +14,26 @@ def run(cmds, fmt = "pandas", spark_master='yarn', app_name='wmfdata', spark_con
 
     if fmt not in ["pandas", "raw"]:
         raise ValueError("The `fmt` should be either `pandas` or `raw`.")
-
     if type(cmds) == str:
         cmds = [cmds]
 
-    result = None
-
     spark_session = spark.get_session(spark_master, app_name, spark_config)
 
-    # TODO figure out how to handle multiple commands
-
-    cmd = cmds[0]
-    spark_result = spark_session.sql(cmd)
-    collected_result = None
-    
+    result_DF = None
+    for cmd in cmds:
+        cmd_result = spark_session.sql(cmd)
+        
+        # If the result has columns, the command was a query and therefore results-producing.
+        # If not, it was a DDL or DML command and not results-producing.
+        if len(cmd_result.columns) > 0:
+            result_DF = cmd_result
     if fmt == 'pandas':
-        collected_result = spark_result.toPandas()
+        result = result_DF.toPandas()
     else:
-        collected_result = spark_result.collect()
+        result = result_DF.collect()
     
     spark.start_session_timeout(spark_session)
-    return collected_result
-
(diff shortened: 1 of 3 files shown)


wmfdata/hive.py
```diff
@@ -14,44 +14,26 @@ def run(cmds, fmt = "pandas", spark_master='yarn', app_name='wmfdata', spark_con
...
+    for cmd in cmds:
+        cmd_result = spark_session.sql(cmd)
+        
+        # If the result has columns, the command was a query and therefore results-producing.
+        # If not, it was a DDL or DML command and not results-producing.
+        if len(cmd_result.columns) > 0:
+            result_DF = cmd_result
     if fmt == 'pandas':
-        return resultDf.toPandas()
+        result = result_DF.toPandas()
```
ottomata, line 31: Won't this overwrite result for each cmd and only return the last result?

wmfdata/spark.py
```diff
@@ -32,5 +69,9 @@ def get_spark_session(master='yarn', app_name='wmfdata', spark_config={}):
     for k, v in spark_config.items():
         builder.config(k, v)
 
-    spark = builder.getOrCreate()
-    return spark
+    session = builder.getOrCreate()
```
ottomata, line 73: This is fine, but for consistency the SparkSession instance is almost always called 'spark'.  It is called this in both the pyspark and scala CLI repls, and is instantiate as such in the Jupyter spark kernels too.

wmfdata/hive.py
nshahquinn: Yes, that's the intention (as described by the function docstring 😊). I added the ability to run multiple commands in the same function call because, when we used impyla, that was the only way to set a session configuration setting before running a query.

wmfdata/spark.py
nshahquinn: Hmm, yeah. I could change this.

Perhaps the bigger issue is that the module itself is called `spark`. I usually start my notebooks with `from wmfdata import hive` or similar, so following that pattern and doing `from wmfdata import spark` could conflict with that convention. Would it be better to call this module `spark_tools`? Or perhaps we could just recommend starting notebooks with `import wmfdata as wmf` and then just using `wmf.spark`?

ottomata: Ah right.  I dunno, maybe just call this spark_session to be clear?

nshahquinn: On reflection, I prefer to leave it as it is. I use `session` consistently throughout this module; adding Spark seems unnecessary because this is the `spark` module (that is also why I renamed `get_spark_session` to `get_session`. When using this module, it will be perfectly possible to do:
```
import wmfdata as wmf

spark = wmf.spark.get_session()
```

That seems clear, concise, and consistent with the wider practice to me. I'll switch to using `wmf.spark`, `wmf.hive`, and so on myself to avoid the potential for issues, and recommend others to do the same.

PR #8, opened 2020-02-26
Support querying Hive tables using multiple engines 

nshahquinn:


wmfdata/hive.py
```diff
@@ -1,40 +1,115 @@
...
     return result
 
+def run(commands, format="pandas", engine="cli"):
+    """
+    Runs SQL commands against the Hive tables in the Data Lake. Currently, this simply passes the commands to the `run_cli` function.
+    """
+
+    if format not in ["pandas", "raw"]:
+        raise ValueError("The `format` should be either `pandas` or `raw`.")
+    if engine not in ("cli"):
```
bearloga: should be `["cli"]`

wmfdata/hive.py
```diff
@@ -1,40 +1,115 @@
...
+    result = None
+
+    # Support multiple commands by concatenating them in one file. If the user has passed more than one query,
+    # this will result in a error when Pandas tries to read the resulting concatenated output (unless the queries
+    # happen to produce the same number of columns).
+    #
+    # Ideally, we would return only the last query's results or throw a clearer error ourselves. However,
+    # there's no simple way to determine if multiple queries have been passed or separate their output,
+    # so it's not worth the effort.
+    merged_commands = ";\n".join(commands)
```
bearloga, line 54: Out of curiosity, do you ever run multiple queries with a single command? Does anyone else? Wondering if this is "well, it might be useful to someone someday" or if there's concrete use cases motivating the inclusion of this feature. Perhaps it would be better to enforce a "one query, one run" rule and leave multiple-querying for the end-user, or, alternatively, move this functionality into, say, `wmfdata.hive.multirun()` to make it explicit.

wmfdata/spark.py
```diff
@@ -43,31 +62,38 @@ def start_session_timeout(session):
     session_timeouts[application_id] = timeout
     timeout.start()
 
-def get_session(master='yarn', app_name='wmfdata', spark_config={}):
+def get_session(type="regular", app_name="wmfdata", extra_settings={}):
     """
     Returns an existent Spark session, or a new one if one hasn't yet been created.
     
     Code calling this is responsible for starting a timeout using `start_session_timeout` when it finishes using the session, in order to prevent idle Spark sessions from wasting cluster resources. This function takes cares of cancelling the timeout if the session is returned again before the timeout finishes.
```
bearloga: In-code documentation should be 80 or 120 characters long (helps with CR)

setup.py
```diff
@@ -11,15 +11,13 @@
     version = metadata["version"],
```
bearloga: Extraneous space, let's remove it as part of this major patch?

wmfdata/hive.py
nshahquinn: Multiple queries in the same function call, no. I've never tried to support that. With this function, multiple queries will result in an error. The other run functions won't error, but they also will only return the last set of results so it's not possible to actually use them to run multiple queries. So, in practice, there is a "one run, one query" rule :)

But multiple _statements_, yes. That has mainly been useful for modifying session settings, which wouldn't persist between two function calls (for example, [increasing the map memory setting](https://wikitech.wikimedia.org/wiki/Analytics/Systems/Cluster/Hive/Queries#Hadoop_containers_run_out_of_memory).) That doesn't necessarily apply to every engine (not to Spark, for example), but I try to keep things as consistent as possible between the run functions.

wmfdata/hive.py
nshahquinn: Ahh, it's not a tuple! I went ahead and changed all these checks to use lists

wmfdata/spark.py
nshahquinn: Ha, yeah, that's pretty nasty. I reformatted all the comments and docstrings in hive.py and spark.py to limit them to 80 characters. I'll do the other files at some point too.

wmfdata/hive.py
bearloga: Why not enforce that the query string you provide must already have all statements in it? e.g. rather than

```python
query = """
SELECT …
"""
commands = ["SET mapred.job.queue.name=nice", "SET mapreduce.map.memory.mb=4096", "USE wmf", query]
wmfdata.hive.run(commands)
```

it would be

```python
query = """
SET mapred.job.queue.name=nice;
SET mapreduce.map.memory.mb=4096;
USE wmf;
SELECT …
"""
wmfdata.hive.run(query)
```

bearloga: OH. For re-reusing queries in multiple engines. My bad!

bearloga (approved):
Everything's great! Thank you for this massive update, Neil! LGTM

PR #11, opened 2020-04-02
Better handle users in different access groups

nshahquinn:


nshahquinn:
This doesn't require a point release now that Luca has deployed [a workaround](https://gerrit.wikimedia.org/r/c/operations/puppet/+/585071), but it's good cleanup of an edge case. Luca took a quick look and gave a thumbs up, but he hasn't worked on this codebase so I still want a review from @bearloga or @nettrom 😊

Also, can whoever merges this use the "rebase and merge" option? It's accessed from the dropdown attached to the "merge pull request" button and will keep the history cleaner by not adding merge commits.

bearloga:
Looks good to me, LGTM

PR #13, opened 2020-08-24
Parameterize spark master location

adamwight:
This allows for a "local" master, short syntax for the number of
cores and so on.

### DIFF
--- a/wmfdata/spark.py
+++ b/wmfdata/spark.py
@@ -66,7 +66,9 @@ def start_session_timeout(session):
     session_timeouts[application_id] = timeout
     timeout.start()
 
-def get_session(type="regular", app_name="wmfdata", extra_settings={}):
+def get_session(
+    type="regular", app_name="wmfdata", master="yarn", extra_settings={}
+):
     """
     Returns an existent Spark session, or a new one if one hasn't yet been
     created.
@@ -89,7 +91,7 @@ def get_session(type="regular", app_name="wmfdata", extra_settings={}):
     # this has happened.
     builder = (
         SparkSession.builder
-        .master("yarn")
+        .master(master)
         .appName(app_name)
         .config(
             "spark.driver.extraJavaOptions",


nshahquinn:
@adamwight thank you for getting involved, and I'm so sorry I haven't reviewed this yet! I'll be on vacation this week, but I'll work on it when I get back on 5 October; I've created [a Phab task](https://phabricator.wikimedia.org/T263954) so I don't forget.

First impressions:
- This seems like a useful feature!
- It seems better to make "local" a new value you can pass as the `type` parameter rather than adding a `master` parameter, to keep the API simpler.
-  We should check with Analytics to see whether there are any other settings we should use for a local session (if I remember correctly @jobar is the Spark expert on the team)
- We should update the docstring to explain when a local session is a good option (I'm not sure myself, but since you wanted to add this, I imagine you have some idea!)

jobar:
> We should check with Analytics to see whether there are any other settings we should use for a local session

The way to parameterize the number of cores for a local session is through the master parameter itself: `local` means 1 core, `local[2]` 2 cores, `local[*]` all cores of the local machine. No need for extra parameterization.
Note: I used the word cores above to mimic the naming used in executors, but the correct word would be threads. In local mode, only the driver runs, and execution of the tasks take place in threads. This also means that in term of memory, only the driver one is used when in local mode.

adamwight:
I would like to refine this patch, but am a bit confused about how to proceed.

Do we really want to default to `spark.executor.cores = 4`, even for a yarn master?  The upstream default is 1 thread, and I'm not sure that more threads per executor makes sense because parallelism is probably across executors, not within each one.

Given my current poor understanding of spark, the docstring I would provide would be along the lines of:
> Leave the default `master="yarn"` when your job should be distributed across all machines in the fleet.  This is the right choice for large data sets.  If your query will deal with small data sets, you can limit to a single machine with `master="local"` (1 thread), or `master="local[4]"` to run in 4 threads.

jobar:
Hi Folks, chiming in to try to help :)
When using spark in local mode, `spark.executor.X` are not taken into account. The way to configure how many threads are uses for computation is by using `local[X]`with X the number of threads. From a memory perspective, local-mode means driver only and therefore it's the `--driver-memory` option that defines how much memory is used.
Hope it helps!

ottomata:
Oh, I didn't see this PR before I made #15.  Mine does a little bit more and adds the ability to construct complete custom SparkSessions.

adamwight:
Closed in favor of #15

PR #14, opened 2020-09-02
dblist parser

adamwight:
Expands symbolic names list "group0" into a list of wiki databases.


nshahquinn:
@adamwight thank you for getting involved, and I'm so sorry I haven't reviewed this yet! I'll be on vacation this week, but I'll work on it when I get back on 5 October; I've created [a Phab task](https://phabricator.wikimedia.org/T263954) so I don't forget.

Initial impressions:
- Definitely seems like a useful feature!
- Should we also update `mariadb.run` so that it automatically expands the group names when `"group0"` or similar is passed as the appropriate parameter?
- I'm inclined to rename this module `wikis` or something similar, because I've been thinking about adding similar functionality that isn't just about `dblists` (e.g. translating database codes to the human-readable wiki names).

adamwight:
>     * Should we also update `mariadb.run` so that it automatically expands the group names when `"group0"` or similar is passed as the appropriate parameter?

That makes sense, thanks for the suggestion.

>     * I'm inclined to rename this module `wikis` or something similar, because I've been thinking about adding similar functionality that isn't just about `dblists` (e.g. translating database codes to the human-readable wiki names).

Agreed!  Now that I think about it, "dblist" would mean something slightly different to many Wikimedia developers, it's important as the mapping between physical DB server and wiki database, but we don't care about that here.

nshahquinn:
@adamwight I just made some tweaks:
- Renamed the function `get_dblist`, which seems a bit more accessible to me
- Moved it into the `utils` module, so that other modules can depend on it. I'm particularly thinking about `mariadb`, although I'm not going to do that right now since I think it requires some more consideration. For example, the "wikipedia" dblist includes various arbcom wikis, which isn't what someone would expect if they run `mariadb.run("FOO", dbs="wikipedia")`.
- Expanded the documentation a bit.

Please look this over and ensure that it makes sense to you and that I haven't added any mistakes (I did test it). If it looks good to you, we can merge it (finally 😛).

adamwight (commented):
Looks right, thanks for fixing up after my drive-by!

nshahquinn:
Great! Thank _you_ for contributing it, and I'm sorry it took so long to get here!

PR #15, opened 2021-01-05
Slight refactor of spark.py to support local and custom SparkSessions

ottomata:
get_session and run remain backwards compatible, except that
get_session now starts the session timer when it is called.

### DIFF
--- a/wmfdata/mariadb.py
+++ b/wmfdata/mariadb.py
@@ -11,6 +11,8 @@ import pandas as pd
 
 from wmfdata.utils import ensure_list, print_err
 
+
+connection=None
 # Close any open connections at exit
 @atexit.register
 def clean_up_connection():
@@ -24,21 +26,21 @@ def connect(db, use_x1=False):
     # database is located on x1.
     if db == "wikishared":
         use_x1 = True
-    
+
     host_command = "analytics-mysql {db} --print-target".format(db=db)
     if use_x1:
         host_command = host_command + " --use-x1"
-    
+
     host = subprocess.run(
-        host_command, 
+        host_command,
         shell=True,
         stdout=subprocess.PIPE,
         universal_newlines=True
     ).stdout.strip().split(":")
-    
+
     if host == ['']:
         raise ValueError("The database '{}' was not found.".format(db))
-    
+
     port = host[1]
     host = host[0]
 
@@ -54,7 +56,7 @@ def connect(db, use_x1=False):
             "Your account does not have permission to access the Analytics "
             "MariaDB cluster."
         )
-        
+
     connection = mysql.connect(
         host=host,
         port=port,
@@ -63,18 +65,18 @@ def connect(db, use_x1=False):
         charset='utf8',
         autocommit=True
     )
-    
+
     return connection
 
 def run_to_pandas(connection, commands, date_col=None, index_col=None):
     result = None
-    
+
     # Specify the MediaWiki date format for each of the date_cols, if any
     if date_col:
         date_col = ensure_list(date_col)
         date_format = "%Y%m%d%H%M%S"
         date_col = {col: date_format for col in date_col}
-    
+
     # To-do: SQL syntax errors cause a chain of multiple Python errors
     # The simplest way to fix this is probably to get the raw results and
     # then turn them into a data frame; this would let us avoid using
@@ -114,7 +116,7 @@ def run(
 ):
     """
     Run SQL queries or commands on the Analytics MediaWiki replicas.
-    
+
     Arguments:
     * `commands`: the SQL to run. A string for a single command or a list of
       strings for multiple commands within the same session (useful for things
@@ -126,7 +128,7 @@ def run(
           `use_x1` is passed)
         * "logs" for the EventLogging
         * "centralauth" for global accounts
-        * "wikishared" for cross-wiki ExtensionStorage 
+        * "wikishared" for cross-wiki ExtensionStorage
         * "staging" for user-writable ad-hoc tests and analysis
     * `use_x1`: whether to the connect to the given database on the
       ExtensionStorage replica (only works for wiki databases or "wikishared").
@@ -142,20 +144,20 @@ def run(
       set a columns or columns as the index. If using raw format, has no
       effect.
     """
-    
+
     # Make single command and database parameters lists
     commands = ensure_list(commands)
     dbs = ensure_list(dbs)
-    
+
     results = []
-        
+
     if format == "pandas":
         for db in dbs:
             connection = connect(db, use_x1)
             result = run_to_pandas(connection, commands, date_col, index_col)
             connection.close()
             results.append(result)
-        
+
         if len(dbs) > 1:
             # Ignore the indexes on the partial results unless a custom index
             # column was designated
@@ -163,11 +165,11 @@ def run(
                 ignore_index = True
             else:
                 ignore_index = False
-                
+
             return pd.concat(results, ignore_index=ignore_index)
         else:
             return results[0]
-    
+
     elif format == "raw":
         for db in dbs:
             connection = connect(db, use_x1)
@@ -178,13 +180,13 @@ def run(
         if len(dbs) > 1:
             # Take the first set of column names since they'll all be the same
             column_names = results[0].column_names
-            
+
             record_sets = [result.records for result in results]
             records = [x for x in chain(record_sets)]
-            
+
             return ResultSet(column_names, records)
         else:
             return results[0]
-    
+
     else:
         raise ValueError("The format you specified is not supported.")
--- a/wmfdata/spark.py
+++ b/wmfdata/spark.py
@@ -6,27 +6,47 @@ from pyspark.sql import SparkSession
 
 from wmfdata.utils import check_kerberos_auth, ensure_list
 
-REGULAR_SPARK_SETTINGS = {
-    "spark.driver.memory": "2g",
-    "spark.dynamicAllocation.maxExecutors": 64,
-    "spark.executor.memory": "8g",
-    "spark.executor.cores": 4,
-    "spark.sql.shuffle.partitions": 256
+"""
+Predefined spark sessions and configs for use with the get_session and run functions.
+"""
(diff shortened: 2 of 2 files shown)


adamwight:
Very nice, it does just what I would hope for.

wmfdata/mariadb.py
```diff
@@ -24,21 +26,21 @@ def connect(db, use_x1=False):
     # database is located on x1.
     if db == "wikishared":
         use_x1 = True
-    
+
```
nshahquinn, line 29: I appreciate all the formatting cleanup, but please do it in a separate pull request next time. That way, it will be easier to review 😊

wmfdata/spark.py
```diff
@@ -6,27 +6,47 @@
...
+# Add previous session 'type' keys for backwards compatibility.
+PREDEFINED_SPARK_SESSIONS['regular'] = PREDEFINED_SPARK_SESSIONS['yarn']
+PREDEFINED_SPARK_SESSIONS['large'] = PREDEFINED_SPARK_SESSIONS['yarn-large']
 
 EXTRA_JAVA_OPTIONS = {
-    "http.proxyHost": "webproxy.eqiad.wmnet",
-    "http.proxyPort": "8080",
-    "https.proxyHost": "webproxy.eqiad.wmnet",
-    "https.proxyPort": "8080"
+    'http.proxyHost': 'webproxy.eqiad.wmnet',
```
nshahquinn: I'm sure it wasn't intentional, but please don't change one type of quote to another for no reason. Similarly, please use double quotes in your new code to be consistent with the existing practice!

wmfdata/spark.py
```diff
@@ -6,27 +6,47 @@
...
+Predefined spark sessions and configs for use with the get_session and run functions.
+"""
+PREDEFINED_SPARK_SESSIONS = {
+    'local': {
+        'master': 'local[2]',  # [2] means use 2 local worker threads
+        'config': {
+            'spark.driver.memory': '4g',
+        }
+    },
+    'yarn': {
```
nshahquinn: Why do you want to change `regular` to `yarn` (and `large` to `yarn-large`)? `get_session` is intended to make it easy to use Spark without needing to understand the internals, so I don't want to ask the user to deal with concepts like YARN. Instead, I want to give them a simple linear scale of computing power, with a sensible default.

Along the same line, I'm inclined to change `local` to `small`, although it might make sense to keep `local` if using a local session has some important difference other than just being less powerful.

wmfdata/spark.py
```diff
@@ -62,24 +90,31 @@ def start_session_timeout(session):
...
-    created.
-    
-    Code calling this is responsible for starting a timeout using
-    `start_session_timeout` when it finishes using the session, in order to
-    prevent idle Spark sessions from wasting cluster resources. This function
-    takes cares of cancelling the timeout if the session is returned again
-    before the timeout finishes.
+
+
+def get_spark_session(
```
nshahquinn: Let's call this something like `get_custom_session` instead. It would be very confusing to have `spark.get_session` be the basic version and `spark.get_spark_session` be the advanced version 😂

wmfdata/spark.py
```diff
@@ -62,24 +90,31 @@ def start_session_timeout(session):
...
+    master='local[2]',
+    app_name='wmfdata',
```
nshahquinn: What about defaulting this to `"wmfdata-custom"` just to provide a little more information in the logs?

wmfdata/spark.py
```diff
@@ -99,28 +134,77 @@ def get_session(type="regular", app_name="wmfdata", extra_settings={}):
...
+    # Apply any provided spark configs.
+    for k, v in spark_config.items():
         builder.config(k, v)
 
     session = builder.getOrCreate()
 
-    # Cancel any existing timeout on the same session, in case an existing
-    # session is being returned
-    cancel_session_timeout(session)
+    if master == 'yarn' and yarn_session_timeout is not None:
```
nshahquinn: The general way you handle timers seems problematic. Because the timer starts when the session is (re)initialized, its period will overlap with the use of the session and could easily end up killing the session while something is running. In my experience, Spark jobs can definitely run longer than an hour.

Moreover, the timeout essentially makes it impossible to run anything that takes longer than an hour, because you can't manually reset the timer while the long-running thing is blocking the kernel! Theoretically, the minority of users who will use `get_custom_session` can adjust the timeout as necessary, but it will be hard to discover the need _and_ hard to do, since it's not clear in advance how long something will take.

Previously, running `get_session` would cancel any timer but _not_ start one, since it had no control over how the session was being used and could not tell when it was inactive (the docstring advised the user to start timeouts manually when appropriate, but of course it's likely that most people won't follow that advice). 

On the other hand, `run` starts a timer when it _finishes_, because it's the user of the session and knows when it becomes inactive.

I'm not immediately sure how best to handle this. Some options:
- Set the default timeout to something much longer, like a day. This would be less likely to kill something in progress, but it could still happen. For example, if a user opens a session, uses it for a while, leaves, and comes back 23 hours later to keep using it, they're back to having an effective timeout of an hour.
- Give up altogether on timers that start when a session is _initialized_. This will lead to more YARN containers staying open for long periods (although this will not affect people who just use the basic `run` interface) but I don't think the effect is any worse than a Jupyter kernel doing the same thing.
- The Spark cluster itself automatically stops the scheduler process if it is inactive for long enough. I'm not sure if it's possible, but it seems like the ideal solution since that way the timer is always based on inactive time rather than time since creation. We'd need to tweak this code so people understand what has happened when they try to use a session that's been stopped, and it could still lead to data loss in some circumstances, but it is much more reasonable to say that a Spark session _at rest_ is not guaranteed to be permanent.

nshahquinn (changes requested):
Thank you for working on this! I'm looking forward to getting this merged and getting to Newpyter 😊

wmfdata/mariadb.py
ottomata: I blame my editor, and I am sorry.

wmfdata/spark.py
ottomata: Done

wmfdata/spark.py
ottomata: There are pretty significant differences, especially regarding to dependencies.  Local has access to your whole conda env, Yarn does not.  

I guess I changed this because 'regular' sounds like it would be 'local' to me. Additionally, the size of a local process itself can be adjusted so, there might be a difference between an e.g. local-large and a yarn-large.

Would it make sense to keep the accurate terms for the predefined sessions, and add aliases like 'large' and 'small'?

wmfdata/spark.py
ottomata: Ok.

wmfdata/spark.py
ottomata: Ah ok, I think I missed some of that.  Alright, let's do your 2nd solution there.  These get session methods will have a timeout param, but it will be disabled by default.  Run will set a timeout.

wmfdata/mariadb.py
ottomata: I also got undefined variable problems without the `connection=None` line.

wmfdata/spark.py
ottomata: Actually, for simplicity, I made it so run is the only thing that starts a timeout.  The others don't bother.

wmfdata/mariadb.py
nshahquinn: > I blame my editor, and I am sorry.

In that case, I'll withdraw the complaint I filed with the code police 😊

> I also got undefined variable problems without the connection=None line.

As I said, I think the cleanup is actually good since it makes it follow PEP-8. Same with this; I was getting those problems too and just hadn't gotten around to fixing it! Just in a separate pull request next time as a kindness to my limited lizard brain.

wmfdata/spark.py
nshahquinn: As I said in chat, this way makes sense to me now. Let's leave it as you've done it. Can you just change "yarn" to "yarn-regular" just so the progression in power is extra clear?

wmfdata/spark.py
nshahquinn: There are still a bunch of single quotes, though?

wmfdata/spark.py
```diff
@@ -130,9 +214,10 @@ def run(commands, format="pandas", session_type="regular", extra_settings={}):
         * "pandas": a Pandas data frame
         * "raw": a list of tuples, as returned by the Spark SQL interface.
     * `session_type`: the type of Spark session to create.
-        * "regular": the default; able to use up to 15% of Hadoop cluster
+        * "local": Run the command in a local Spark process.
```
nshahquinn: Can you add the one sentence explanation of when to use a local session that you gave me? Also in the `get_session` docstring.

wmfdata/spark.py
```diff
@@ -99,28 +134,77 @@ def get_session(type="regular", app_name="wmfdata", extra_settings={}):
...
+):
+    """
+    Returns a Spark session based on one of the PREDEFINED_SPARK_SESSION types.
+
+    If running in Yarn, the returned SparkSession will have a timeout of 5 minutes.
+    Call session.restart_timeout() to refresh this timeout.
+    Using this or the run function below will also refresh the timeout every time it is called.
+
+    Arguments:
+    * `session_type`: the type of Spark session to create.
```
nshahquinn: This is called `type` in this function.

nshahquinn (changes requested):
It's looking great! I requested a few small tweaks; otherwise it's ready to merge.

wmfdata/spark.py
ottomata: Uh, I thought I fixed that.  OooK.

wmfdata/spark.py
ottomata: Done.

wmfdata/spark.py
ottomata: Done.

nshahquinn (approved):
Thank you so much! 😊

PR #17, opened 2021-01-29
Fix various minor issues

nshahquinn:
It will be easier to review this commit-by-commit instead of reviewing all the changes at once.

When merging, we should _not_ squash these all together since they deal with various separate issues.


bearloga:
Excellent changes, thank you! LGTM

PR #18, opened 2021-02-01
Add Presto support

nshahquinn:
This is based on the `misc_cleanup` branch, so it shouldn't be merged until #17 is.

The only commit to review here is the last (5fcfd0b).

### DIFF
--- a/wmfdata/presto.py
+++ b/wmfdata/presto.py
@@ -1,0 +1,86 @@
+import os
+import warnings
+
+import pandas as pd
+import prestodb
+from requests.packages import urllib3
+
+from wmfdata.utils import (
+    check_kerberos_auth,
+    ensure_list
+)
+
+# Disable a warning issued by urllib3 because the certificate for an-coord1001.eqiad.wmnet
+# does not specify the hostname in the subjectAltName field (T158757)
+SubjectAltNameWarning = urllib3.exceptions.SubjectAltNameWarning
+urllib3.disable_warnings(SubjectAltNameWarning)
+
+def run(commands, catalog="analytics_hive"):
+    """
+    Runs one or more SQL commands using the Presto SQL engine and returns the last result
+    in a Pandas DataFrame.
+    
+    Presto can be connected to many different backend data stores, or catalogs.
+    Currently it is only connected to the Data Lake, with has the catalog name "analytics_hive".
+
+    """
+    commands = ensure_list(commands)
+    check_kerberos_auth()
+
+    USER_NAME = os.getenv("USER")
+    PRESTO_AUTH = prestodb.auth.KerberosAuthentication(
+        config="/etc/krb5.conf",
+        service_name="presto",
+        principal=f"{USER_NAME}@WIKIMEDIA",
+        ca_bundle="/etc/ssl/certs/Puppet_Internal_CA.pem"
+    )
+
+    connection = prestodb.dbapi.connect(
+        catalog=catalog,
+        host="an-coord1001.eqiad.wmnet",
+        port=8281,
+        http_scheme="https",
+        user=USER_NAME,
+        auth=PRESTO_AUTH,
+        source=f"{USER_NAME}, wmfdata-python"
+    )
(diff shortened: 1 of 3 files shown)


wmfdata/presto.py
```diff
@@ -0,0 +1,80 @@
...
+    PRESTO_AUTH = prestodb.auth.KerberosAuthentication(
+        config="/etc/krb5.conf",
+        service_name="presto",
+        principal=f"{USER_NAME}@WIKIMEDIA",
+        ca_bundle="/etc/ssl/certs/Puppet_Internal_CA.pem"
+    )
+
+    connection = prestodb.dbapi.connect(
+        catalog=catalog,
+        # This should be "analytics-hive.eqiad.wmnet", but doing that gives us cert errors
```
ottomata: Hm, interesting.  Ping @elukey

ottomata:
Interesting, LGTM.  Too bad about that analytics-hive.eqiad.wmnet thing; when we have to replace an-coord1001, or fail it over, clients using this lib will break, eh?  Maybe @elukey can help us

elukey:
> Interesting, LGTM. Too bad about that analytics-hive.eqiad.wmnet thing; when we have to replace an-coord1001, or fail it over, clients using this lib will break, eh? Maybe @elukey can help us

In this case it is a little more complicated than the analytics-hive use case, since TLS is also involved. The high level things to do are the following (in my opinion):
* Create a DNS CNAME analytics-presto.eqiad.wmnet pointing to an-coord1001
* Create a TLS certificate (cergen?) for analytics-presto.eqiad.wmnet and force the Presto Query coordinator to use it, instead the host-level one that it is currently in use (an-coord1001.eqiad.wmnet).
* Add `presto/analytics-presto.eqiad.wmnet@WIKIMEDIA` to the presto keytab for an-coord1001, and then create a similar one for an-coord1002
* Deploy a presto coordinator on an-coord1002
* Update the Presto workers config to use analytics-presto.eqiad.wmnet to report their statuses.

After the above steps we should be able to failover transparently with Presto (like we do for Hive). Does it make sense?
In theory the above should work, and it should be doable in a reasonable amount of time. I can try to test it during the following days, would it be ok? Lemme know :)

ottomata:
Sounds great, I'll create a Phab ticket.

ottomata:
https://phabricator.wikimedia.org/T273642

ottomata:
@neilpquinn I think you should merge this as is, and when we get an HA presto name to use, we can switch this then.

nshahquinn:
> @neilpquinn I think you should merge this as is, and when we get an HA presto name to use, we can switch this then.

Makes sense!

One related point: when I run `presto.run` with `an-coord1001.eqiad.wmnet` as the coordinator, I get this certificate warning:

> /home/neilpquinn-wmf/venv/lib/python3.7/site-packages/urllib3/connection.py:384: SubjectAltNameWarning: Certificate for an-coord1001.eqiad.wmnet has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/urllib3/urllib3/issues/497 for details.)
>   SubjectAltNameWarning,

Should we update the cert, or should I just suppress it?

ottomata:
> Should we update the cert, or should I just suppress it?

Both?  Actually, I'm not totally following what this error is from.  The only SSL I can see in the path is for Kerberos auth, which is using the puppet CA cert, which we are unlikely to ever change.  Again, maybe @elukey has some context.

But, I think you can suppress that error for now.

elukey:
For the moment I think that we can keep it in that way, see https://phabricator.wikimedia.org/T158757

If we create analytics-presto.eqiad.wmnet we'll need to create another cert anyway so we can make sure that this is tracked in https://phabricator.wikimedia.org/T273642

nshahquinn:
Thanks @ottomata and @elukey! I've suppressed that particular warning for now, so I'll go ahead and merge.

PR #19, opened 2021-02-22
Update docs to reflect new `run` signature

adamwight:


adamwight:
I'm running into other deprecations...  Is there no longer an equivalent to `mariadb.list_wikis`?

adamwight:
> I'm running into other deprecations... Is there no longer an equivalent to `mariadb.list_wikis`?

Oops, I was pulling from my outdated fork rather than this repo.  I see now that the dblist change landed!  I'll add to this PR...

nshahquinn:
Thank you very much!

It looks like you may have been planning to add something else but hopefully the happiness of seeing "merged" will cancel out the sadness of having to open another PR 😂

PR #20, opened 2021-02-22
Missing comma breaks install

adamwight:


nshahquinn:
Thank you for noticing that and jumping in with a PR! I actually noticed and fixed it independently before I saw this; otherwise, I would definitely have merged it.

PR #21, opened 2021-02-23
Timeout remote version checking

milimetric:
In places where this remote checking can't work, like restricted
networks, it will attempt until the connection is closed, which can take
minutes.  In the vast majority of cases it shouldn't take more than 1
second, so setting that timeout to let it fail fast in those cases.


nshahquinn:
Excellent point! Thanks, Dan 😊

PR #22, opened 2021-03-01
Add spark support for shipping conda environments and DRY http proxy setting propagation

ottomata:
This helps automate shipping of conda environments
to remote YARN workers when there are dependencies
needed there that are not installed there by default.

This also properly sets up PYSPARK_PYTHON to use
anaconda-wmf on remote YARN workers if the user
chooses not to ship a conda environment of their own.
anaconda-wmf is installed everywhere and is a better default
for remote python than the system installed one.

Bug: T272313

### DIFF
--- a/wmfdata/conda.py
+++ b/wmfdata/conda.py
@@ -8,22 +8,22 @@ try:
 except ImportError:
     conda_installed = False
 
-from wmfdata.utils import log
+from wmfdata.utils import print_err
 
 """
 Default CONDA_BASE_ENV_PREFIX.
 """
-conda_base_env_prefix_default = '/usr/lib/anaconda-wmf'
+conda_base_env_prefix_default = "/usr/lib/anaconda-wmf"
 
 """
 Default kwargs to pass to conda_pack.pack.
 """
 conda_pack_defaults = {
     # This is required to pack stacked conda envs.
-    'ignore_editable_packages': True,
-    'verbose': True,
-    'force': False,
-    'format': 'tgz',
+    "ignore_editable_packages": True,
+    "verbose": True,
+    "force": False,
+    "format": "tgz",
 }
 
 def info():
@@ -31,7 +31,7 @@ def info():
     Returns the result of conda info --all --json as a dict.
     """
     if conda_installed:
-        return json.loads(condacli('info', ["--all", "--json"])[0])
+        return json.loads(condacli("info", ["--all", "--json"])[0])
     else:
         return {}
 
@@ -39,26 +39,26 @@ def env_vars():
     """
     Returns dict of conda environment variables.
     """
-    return info().get('env_vars', {})
+    return info().get("env_vars", {})
 
 def active_prefix():
     """
     Returns the path to the active conda env, or None.
     """
-    return info().get('active_prefix', None)
+    return info().get("active_prefix", None)
 
 def active_name():
     """
     Returns the name of the active conda env, or None.
     """
-    return env_vars().get('CONDA_DEFAULT_ENV', None)
+    return env_vars().get("CONDA_DEFAULT_ENV", None)
 
 def base_prefix():
     """
     If the active conda env is stacked on another conda env, returns
     the path to that base conda env, or None.
     """
-    return env_vars().get('CONDA_PREFIX_1', None)
+    return env_vars().get("CONDA_PREFIX_1", None)
 
 def is_active():
     """
@@ -84,14 +84,14 @@ def conda_base_env_prefix():
     Returns the path to the conda base env, which on WMF servers is /usr/lib/anaconda-wmf.
     This can be overridden by setting the CONDA_BASE_ENV_PREFIX env var.
     """
-    return os.environ.get('CONDA_BASE_ENV_PREFIX', conda_base_env_prefix_default)
+    return os.environ.get("CONDA_BASE_ENV_PREFIX", conda_base_env_prefix_default)
 
 
 def pack(**conda_pack_kwargs):
     """
     Calls conda_pack.pack.
     If the packed output file already exists, this will not repackage
-    it unless conda_pack_kwargs['force'] == True.
+    it unless conda_pack_kwargs["force"] == True.
 
     Returns the path to the output packed env file.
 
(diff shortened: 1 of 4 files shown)


wmfdata/conda.py
```diff
@@ -0,0 +1,118 @@
...
+        return False
+    else:
+        return True
+
+def conda_base_env_path():
+    """
+    Returns the path to the conda base env, which on WMF servers is /usr/lib/anaconda-wmf.
+    This can be overridden by setting the CONDA_BASE_ENV_PATH env var.
+    """
+    return os.environ.get('CONDA_BASE_ENV_PATH', '/usr/lib/anaconda-wmf')
```
declerambaul: make this a variable on top to avoid a hard coded path?

wmfdata/conda.py
```diff
@@ -0,0 +1,118 @@
...
+    kwargs.update(conda_pack_kwargs)
+
+    # Make sure output is set to something, so we can return it if it already exists.
+    if 'output' not in kwargs:
+        conda_env_name = 'conda-env'
+        if 'prefix' in kwargs:
+            conda_env_name = os.path.basename(kwargs['prefix'])
+        elif 'name' in kwargs:
+            conda_env_name = kwargs['name']
+        elif is_active():
```
declerambaul, line 111: what happens if there isn't an active conda env? will there be a `conda-conda-env.tgz` file or do we expect this to fail?

wmfdata/conda.py
```diff
@@ -0,0 +1,118 @@
...
+            conda_env_name = active_name()
+
+        kwargs['output'] = 'conda-{}.{}'.format(conda_env_name, kwargs['format'])
+
+    conda_packed_file = kwargs['output']
+    if os.path.isfile(conda_packed_file) and not kwargs['force']:
+        log.info(
+            f'A conda environment is already packed at {conda_packed_file}. '
```
declerambaul: specify that this already packed env will be shipped to the cluster

declerambaul (approved):
Nice! 

I tested with a simple example

```

! pip install pillow
! pip install --ignore-installed --upgrade git+https://github.com/ottomata/wmfdata.git@conda
import wmfdata
spark =wmfdata.spark.get_session(ship_python_env=True)
df = spark.createDataFrame(data=[('no.jpg',)], schema=['f'])

@F.udf(returnType='string')
def fakeit(f):
    from PIL import Image
    try:
        Image.open(f)
    except Exception as e:
        return f'{e}'
    return f

df = df.withColumn('t', fakeit(F.col('f')))
df.toPandas()

# 0	no.jpg	[Errno 2] No such file or directory: 'no.jpg'
```

ottomata:
Wow actually, Luca just sent me:

https://github.com/criteo/cluster-pack

Which looks maybe way nicer than my patch here.  Let me try that.

ottomata:
> https://github.com/criteo/cluster-pack

Looked through code, and while it does a lot more stuff, essentially what it is doing is exactly what this patch does.  cluster-pack looks like it should be able to detect if new dependencies are installed and only re-pack and upload if so.  This patch makes you do that manually (by passing a force=True option).

As is, cluster-pack won't work, as our stacked conda setup makes it so we need to pass `ignore_editable_packages=True` to conda_pack, and cluster-pack doesn't currently have an option to pass that down.

Let's proceed with this patch.

wmfdata/conda.py
ottomata: Oo thanks for noticing the conda-conda thing.  Fixed.  but yes, the default output should be conda-env.tgz.  If no environment is active and `conda_pack_kwargs` aren't enough for `conda_pack.pack` to figure out what to pack, this will error.  I'll add a comment.

wmfdata/conda.py
ottomata: Hm, it won't necessarily at this point.  This `conda.pack` function is just for packing.  Uploading to the cluster is done in spark.py by setting `spark.yarn.dist.archives`.  There is a log message there that specifies that this will be shipped to the cluster.

wmfdata/spark.py
```diff
@@ -1,15 +1,22 @@
...
 
 import findspark
-findspark.init("/usr/lib/spark2")
-from pyspark.sql import SparkSession
 
+from wmfdata import conda
 from wmfdata.utils import (
     check_kerberos_auth,
-    ensure_list
+    ensure_list,
```
nshahquinn, line 10: Please alphabetize the list :)

wmfdata/conda.py
```diff
@@ -0,0 +1,131 @@
...
+    conda_installed = True
+except ImportError:
+    conda_installed = False
+
+from wmfdata.utils import log
+
+"""
+Default CONDA_BASE_ENV_PREFIX.
+"""
+conda_base_env_prefix_default = '/usr/lib/anaconda-wmf'
```
nshahquinn: Please use double quotes throughout this file :)

wmfdata/utils.py
```diff
@@ -9,6 +10,19 @@
 import pandas as pd
 import requests
 
+"""
+Global wmfdata.utils.log logger.  Usage:
+
+from wmfdata.utils import log
+log.info("my message")
+"""
+log = logging.getLogger('wmfdata')
```
nshahquinn: Double quotes.

wmfdata/utils.py
```diff
@@ -9,6 +10,19 @@
...
+log.setLevel(logging.INFO)
+formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s %(name)s: %(message)s', datefmt='%Y-%m-%dT%H:%M:%S')
+handler = logging.StreamHandler(sys.stdout)
```
nshahquinn: I'd use `sys.stderr` instead. In a Jupyter notebook, that gets displayed with a light red background, so it's clearly distinguishable as a log message. The downside is the red makes it seem more like a warning, but to me it's worth it (I use the print_err function below for a number of informational things).

wmfdata/utils.py
```diff
@@ -9,6 +10,19 @@
...
 
+"""
+Global wmfdata.utils.log logger.  Usage:
+
+from wmfdata.utils import log
+log.info("my message")
+"""
+log = logging.getLogger('wmfdata')
+log.setLevel(logging.INFO)
+formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s %(name)s: %(message)s', datefmt='%Y-%m-%dT%H:%M:%S')
```
nshahquinn: Is the datetime really necessary? It makes the messages quite long. It's particularly unnecessary in Jupyter since there the messages are clearly tied to the cell that generated them.

wmfdata/spark.py
```diff
@@ -1,15 +1,22 @@
 from threading import Timer
 import warnings
+import os
 
 import findspark
-findspark.init("/usr/lib/spark2")
-from pyspark.sql import SparkSession
 
+from wmfdata import conda
```
nshahquinn, line 7: The top of init.py says that modules can depend only depend on utils to prevent circular import issues. Can you update that to say depending on conda is okay too?

Also, you may want to import conda there too so it's available after `import wmfdata` like all the other modules. But I doubt it's going to be used by any end users, so perhaps we should just leave it out? Up to you.

wmfdata/spark.py
```diff
@@ -56,7 +63,9 @@
 def get_custom_session(
     master="local[2]",
     app_name="wmfdata-custom",
-    spark_config={}
+    spark_config={},
+    ship_python_env = False,
```
nshahquinn: You've got some extra space around the `=` here and in the line below :)

nshahquinn (commented):
Okay, I've taken a look and overall it seems good!

I read through all the conda bits, so I understand them at a basic level, but not enough to actually comment on the details. I feel comfortable trusting you, though :)

I did leave some minor comments about code formatting and the logging. Once you deal with those, feel free merge it yourself.

nshahquinn:
Also, thanks for testing this, @declerambaul! It made this review a lot easier 😊

wmfdata/conda.py
ottomata: Thanks, hard habit to break!

wmfdata/utils.py
ottomata: To keep this PR simple, I'll just use print_err instead of a logger then.

We can add better logging in a future PR.

wmfdata/utils.py
ottomata: Using print_err

wmfdata/spark.py
ottomata: ok. Lemme know if I did that right...

PR #23, opened 2021-10-21
Switch to PyHive

milimetric:
See the notes for the pyhive commit for testing info.

Note: there's some timeout small change from the past... I guess I forgot to send it, I'm not sure if it's relevant anymore

PR #24, opened 2021-12-07
Switch to pyhive as a hive interface

milimetric:
To test your notebook's conda environment needs some help:
  conda install -c conda-forge cyrus-sasl
  export CPPFLAGS="${CPPFLAGS} -isystem ${CONDA_PREFIX}/include"

In production, we would apply these changes to the base conda
environment.  There's a discussion on the need for this in phabricator.

(it needs this to allow `pip install pyhive[hive]`)

Bug: T275233

### DIFF
--- a/wmfdata/hive.py
+++ b/wmfdata/hive.py
@@ -7,6 +7,7 @@ import warnings
 from pyhive import hive
 from shutil import copyfileobj
 from wmfdata.utils import (
+    check_kerberos_auth,
     ensure_list
 )
 
@@ -15,7 +16,7 @@ KERBEROS_SERVICE_NAME = "hive"
 
 
 def run_cli(
-    commands, format="pandas", heap_size=1024, use_nice=True,
+    commands, format="pandas", heap_size=None, use_nice=True,
     use_ionice=True
 ):
     """
@@ -30,19 +31,19 @@ def run_cli(
     * `format`: what format to return the results in
         * "pandas": a Pandas data frame
         * "raw": a TSV string, as returned by the command line interface.
-    * `heap_size`: the amount of memory available to the Hive client. Increase
-      this if a command experiences an out of memory error.
+    * `heap_size`: [Deprecated]
     * `use_nice`: [Deprecated]
     * `use_ionice`: [Deprecated]
     """
     warnings.warn(
-        "'run_cli' is deprecated. It will be removed in the next major release.",
+        "'run_cli' is deprecated. It will be removed in the next major release."
+        "Please use 'run' instead.",
         category=FutureWarning
     )
-    return run(commands, format, heap_size)
+    return run(commands, format)
 
 
-def run(commands, format="pandas", heap_size=1024, engine="deprecated"):
+def run(commands, format="pandas", heap_size="deprecated", engine="deprecated"):
     """
     Runs SQL commands against the Hive tables in the Data Lake.
 
@@ -54,16 +55,16 @@ def run(commands, format="pandas", heap_size=1024, engine="deprecated"):
     * `format`: what format to return the results in
         * "pandas": a Pandas data frame
         * "raw": a TSV string, as returned by the command line interface.
-    * `heap_size`: the amount of memory available to the Hive client. Increase
-      this if a command experiences an out of memory error.
+    * `heap_size`: [Deprecated]
+    * `engine`: [Deprecated]
     """
 
     if format not in ["pandas", "raw"]:
         raise ValueError("The `format` should be either `pandas` or `raw`.")
 
-    if heap_size != 1024:
+    if heap_size != "deprecated":
         warnings.warn(
-            "'heap_size' is not being applied. TODO: can pyhive set this somehow?",
+            "'heap_size' is deprecated. It will be removed in the next major release",
             category=FutureWarning
         )
 
@@ -72,6 +73,8 @@ def run(commands, format="pandas", heap_size=1024, engine="deprecated"):
             "'engine' is deprecated. It will be removed in the next major release.",
             category=FutureWarning
         )
+
+    check_kerberos_auth()
 
     connect_kwargs = {
         "host": HIVE_URL,
--- a/wmfdata/hive.py
+++ b/wmfdata/hive.py
@@ -1,25 +1,25 @@
-import datetime as dt
 import os
-import re
-from shutil import copyfileobj
-import subprocess
+import pwd
 import tempfile
-import warnings
-
 import pandas as pd
 
+from pyhive import hive
+from shutil import copyfileobj
 from wmfdata.utils import (
-    check_kerberos_auth,
     ensure_list
 )
 
+HIVE_URL = 'analytics-hive.eqiad.wmnet'
+KERBEROS_SERVICE_NAME = 'hive'
+
+
 def run_cli(
-  commands, format = "pandas", heap_size = 1024, use_nice = True,
-  use_ionice = True
+    commands, format="pandas", heap_size=1024, use_nice=True,
+    use_ionice=True
 ):
     """
-    Runs SQL commands against the Hive tables in the Data Lake using Hive's
-    command line interface.
+    [DEPRECATED] Formerly ran SQL commands against the Hive tables in the Data Lake
+    using Hive's command line interface.  Currently just calls `run`.
 
     Arguments:
     * `commands`: the SQL to run. A string for a single command or a list of
@@ -31,30 +31,32 @@ def run_cli(
         * "raw": a TSV string, as returned by the command line interface.
     * `heap_size`: the amount of memory available to the Hive client. Increase
       this if a command experiences an out of memory error.
-    * `use_nice`: Run with a lower priority for processor usage.
-    * `use_ionice`: Run with a lower priority for disk access.
+    * `use_nice`: [Deprecated]
+    * `use_ionice`: [Deprecated]
+    """
+    return run(commands, format, heap_size)
+
+
+def run(commands, format="pandas", heap_size=1024):
+    """
+    Runs SQL commands against the Hive tables in the Data Lake.
+
+    Arguments:
+    * `commands`: the SQL to run. A string for a single command or a list of
+      strings for multiple commands within the same session (useful for things
+      like setting session variables). Passing more than one query is *not*
+      supported, and will usually result in an error.
+    * `format`: what format to return the results in
+        * "pandas": a Pandas data frame
+        * "raw": a TSV string, as returned by the command line interface.
+    * `heap_size`: the amount of memory available to the Hive client. Increase
+      this if a command experiences an out of memory error.
     """
 
-    commands = ensure_list(commands)
     if format not in ["pandas", "raw"]:
-        raise ValueError("'{}' is not a valid format.".format(format))
-    if format == "raw":
-        warnings.warn(
-            "The 'raw' format is deprecated. It will be removed in the next major release.",
-            category=FutureWarning
-        )
-    check_kerberos_auth()
+        raise ValueError("The `format` should be either `pandas` or `raw`.")
 
-    shell_command = "export HADOOP_HEAPSIZE={0} && "
-    if use_nice:
-        shell_command += "/usr/bin/nice "
-    if use_ionice:
-        shell_command += "/usr/bin/ionice "
-    shell_command += "/usr/bin/hive -S -f {1}"
-
-    result = None
-
-    # Support multiple commands by concatenating them in one file. If the user
+    # Support multiple commands by concatenating them with ";". If the user
     # has passed more than one query, this will result in a error when Pandas
     # tries to read the resulting concatenated output (unless the queries
     # happen to produce the same number of columns).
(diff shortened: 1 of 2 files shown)


milimetric:
Uh... I guess #23 was automatically closed, sorry @nshahquinn, adding you back here.  I put the clarification about cyrus-sasl in the commit message.

wmfdata/hive.py
```diff
@@ -1,25 +1,45 @@
...
 def run_cli(
-  commands, format = "pandas", heap_size = 1024, use_nice = True,
-  use_ionice = True
+    commands, format="pandas", heap_size=1024, use_nice=True,
+    use_ionice=True
 ):
     """
-    Runs SQL commands against the Hive tables in the Data Lake using Hive's
-    command line interface.
+    [DEPRECATED] Formerly ran SQL commands against the Hive tables in the Data Lake
```
nshahquinn, line 24: We should also log a message about the deprecation if this function is used, similar to https://github.com/wikimedia/wmfdata-python/blob/master/wmfdata/mariadb.py#L148

Also, can you update the `load_csv` function to use `run` instead of `run_cli`, on lines 144 and 145 below?

wmfdata/hive.py
```diff
@@ -63,77 +65,23 @@ def run_cli(
     # error ourselves. However, there's no simple way to determine if multiple
     # queries have been passed or separate their output, so it's not worth
     # the effort.
-    merged_commands = ";\n".join(commands)
+    commands = ";\n".join(ensure_list(commands))
+
+    connect_kwargs = {
+        'host': HIVE_URL,
```
nshahquinn: Please use double quotes, just to stay consistent 😊

wmfdata/hive.py
```diff
@@ -1,25 +1,45 @@
...
+        * "raw": a TSV string, as returned by the command line interface.
+    * `heap_size`: the amount of memory available to the Hive client. Increase
+      this if a command experiences an out of memory error.
+    * `use_nice`: [Deprecated]
+    * `use_ionice`: [Deprecated]
+    """
+    return run(commands, format, heap_size)
+
+
+def run(commands, format="pandas", heap_size=1024):
```
nshahquinn: It looks like the `heap_size` argument isn't used. If it's not useful to implement that feature (I'm not sure), we should mark it as deprecated and log a warning.

We should also keep the `engine` parameter, with a deprecation warning.

wmfdata/hive.py
```diff
@@ -63,77 +65,23 @@ def run_cli(
     # error ourselves. However, there's no simple way to determine if multiple
     # queries have been passed or separate their output, so it's not worth
     # the effort.
-    merged_commands = ";\n".join(commands)
+    commands = ";\n".join(ensure_list(commands))
```
nshahquinn: It looks like you can't execute multiple commands this way when using a Python client. Instead, you'll want to loop through the commands, saving the output to a variable if it exists. Here's how we do it with Spark: https://github.com/wikimedia/wmfdata-python/blob/master/wmfdata/spark.py#L316

nshahquinn (changes requested):

wmfdata/hive.py
```diff
@@ -31,30 +51,12 @@ def run_cli(
...
 
-    commands = ensure_list(commands)
     if format not in ["pandas", "raw"]:
-        raise ValueError("'{}' is not a valid format.".format(format))
-    if format == "raw":
-        warnings.warn(
-            "The 'raw' format is deprecated. It will be removed in the next major release.",
-            category=FutureWarning
-        )
-    check_kerberos_auth()
```
nshahquinn, line 46: Make sure to use this check in the `run` function.

wmfdata/hive.py
```diff
@@ -31,30 +51,12 @@ def run_cli(
...
     * `heap_size`: the amount of memory available to the Hive client. Increase
       this if a command experiences an out of memory error.
-    * `use_nice`: Run with a lower priority for processor usage.
-    * `use_ionice`: Run with a lower priority for disk access.
     """
 
-    commands = ensure_list(commands)
     if format not in ["pandas", "raw"]:
-        raise ValueError("'{}' is not a valid format.".format(format))
-    if format == "raw":
```
nshahquinn, line 41: Could you move this deprecation warning to the `run` function?

nshahquinn (changes requested):

nshahquinn:
> To test your notebook's conda environment needs some help:
conda install -c conda-forge cyrus-sasl
export CPPFLAGS="${CPPFLAGS} -isystem ${CONDA_PREFIX}/include"
>
> In production, we would apply these changes to the base conda
environment. There's a discussion on the need for this in phabricator.

That makes sense! Can you point me to the discussion on Phabricator? I don't see it in [T275233](https://phabricator.wikimedia.org/T275233). We should probably get that done before merging this to the main branch, and definitely before releasing this in a new version.

wmfdata/hive.py
milimetric: it doesn't seem to be necessary, pyhive seems to check it and print a proper error, do you think it's still useful?

wmfdata/hive.py
milimetric: but "raw" isn't deprecated, it still works, no?

wmfdata/hive.py
milimetric: hm, so that just returns the results from the last query or None.  Would we maybe just want to return an array of all the results formatted as requested?

wmfdata/hive.py
milimetric: > Can you point me to the discussion on Phabricator? I don't see it in [T275233](https://phabricator.wikimedia.org/T275233). We should probably get that done before merging this to the main branch, and definitely before releasing this in a new version.

I tried to find it, and couldn't, I'm sorry.  Luca pointed me to it, I'll ask him.  But more importantly, yeah, let's get Andrew or someone to just set up the environment.

wmfdata/hive.py
milimetric: oh, I saw single quotes in a few places and wasn't sure.  Double quotes is fine, done.

wmfdata/hive.py
milimetric: I couldn't figure out how to set the heap_size through pyhive, I think I'm missing something fundamental.  I asked for help and nobody else knew, but I don't think we should deprecate it.  I'll just add a warning that we're not setting it.

(done on engine)

milimetric:
> Can you point me to the discussion on Phabricator? I don't see it in T275233. We should probably get that done before merging this to the main branch, and definitely before releasing this in a new version.

I tried to find it, and couldn't, I'm sorry. Luca pointed me to it, I'll ask him. But more importantly, yeah, let's get Andrew or someone to just set up the environment.

wmfdata/hive.py
milimetric: done, and done

wmfdata/hive.py
milimetric: EDIT: yep, definitely failing, nvm my previous message:

(not true! :) ---> And I tested multiple commands, they seemed to work ok with my concatenation, did I miss something?  What commands failed for you?)

wmfdata/hive.py
nshahquinn: Thanks! It would also be good if the deprecation message advises using `run` instead.

wmfdata/hive.py
nshahquinn: Well, I take "deprecated" to mean that it's recommended not to be used. And the existing code has a recommendation not to use the raw format (as with the other run functions). So I want to make sure that continues to be the case.

If you think we should _keep_ the raw format, that's a valid conversation, but a separate one 😊

wmfdata/hive.py
nshahquinn: Hmm, what did you run?

I tried running 
```
import wmfdata as wmf

hive_db = "nshahquinn"
wmf.hive.run([
    f"DROP TABLE IF EXISTS {hive_db}.wmfdata_test_1",
    f"DROP TABLE IF EXISTS {hive_db}.wmfdata_test_2",
    f"DROP TABLE IF EXISTS {hive_db}.wmfdata_test_empty",
])
```

That gave me the following error:
> Error while compiling statement: FAILED: ParseException line 1:46 missing EOF at ';' near 'wmfdata_test_1'

wmfdata/hive.py
nshahquinn: Well, the error we throw is much clearer and more helpful to the user:
> OSError: You do not have Kerberos credentials. Authenticate using `kinit` or run your script as a keytab-enabled user.

Compare that to the default error:
> TTransportException: Could not start SASL: b'Error in sasl_client_start (-1) SASL(-1): generic failure: GSSAPI Error: No credentials were supplied, or the credentials were unavailable or inaccessible (No Kerberos credentials available (default cache: FILE:/tmp/krb5cc_12049))'

Besides that, we use this function with all the run functions, so keeping it makes things consistent and gives us a single function we can modify later if the Kerberos setup changes.

wmfdata/hive.py
nshahquinn: Hmm. I did some searching and I'm starting to think that setting the heap size isn't relevant when we're not using the `hive` CLI. That setting controls the memory used by a Hadoop process, so setting it on a worker in the Hadoop cluster always makes sense. But in this case, I think it was actually controlling the memory used by the _local_ Hadoop process started by the `hive` CLI, which was a thick client. But with `beeline` and PyHive, I don't think any processing actually happens on the client, which would mean there are no client-side memory settings to configure.

So that's my suspicion, which would mean it makes sense to deprecate the option. But I definitely could be misunderstanding the architecture.

nshahquinn:
> > Can you point me to the discussion on Phabricator? I don't see it in T275233. We should probably get that done before merging this to the main branch, and definitely before releasing this in a new version.
> 
> I tried to find it, and couldn't, I'm sorry. Luca pointed me to it, I'll ask him. But more importantly, yeah, let's get Andrew or someone to just set up the environment.

Is [T292699](https://phabricator.wikimedia.org/T292699) the task for the necessary `CPPFLAGS` change?

wmfdata/hive.py
nshahquinn: Oh, I see, you updated your message to say it _was_ failing. When I replied yesterday, I thought you were saying the opposite :joy:

wmfdata/hive.py
milimetric: You can still control the memory used by the cluster, I think.  So I was thinking of finding a way to do that, but you're right that heap_size wouldn't be the right name for that anyway.  You're right, deprecating.

wmfdata/hive.py
milimetric: my bad, yeah, this is now fixed

milimetric:
> Is T292699 the task for the necessary CPPFLAGS change?

Yes!  I'm following up on it now.

nshahquinn:
@milimetric thanks for the follow up! I see two issues still outstanding:
1. There's one comment still open about reinstating the deprecation notice on the "raw" format.
2. I just tried installing this version in a new environment, and I got an error where Pip was unable to build the Sasl module ([pip output](https://phabricator.wikimedia.org/P19375)). This happened whether or not Cyrus-Sasl was installed. Any ideas what's going on? (And, related to that, is Cyrus-Sasl still a required dependency? Because we haven't made any provision for getting it installed yet.)

milimetric:
> 1. There's one comment still open about reinstating the deprecation notice on the "raw" format.

I'll take care of that soon, sorry, been out sick

> 2. I just tried installing this version in a new environment, and I got an error where Pip was unable to build the Sasl module ([pip output](https://phabricator.wikimedia.org/P19375)). This happened whether or not Cyrus-Sasl was installed. Any ideas what's going on? (And, related to that, is Cyrus-Sasl still a required dependency? Because we haven't made any provision for getting it installed yet.)

@ottomata what do you think?  In theory just installing cyrus-sasl before was all I needed.

wmfdata/hive.py
milimetric: I see.  I am not a user so I don't know, I'd imagine raw would be useful if you didn't want to get pandas involved, but maybe that's just always better.  I noticed somehow `pd.read_sql` from below seems faster than the raw fetch.  If it's really going to be removed in the next major release, then totally, let's deprecate (done)

nshahquinn: I see what you mean. I'm going to resolve this comment and move the discussion to https://phabricator.wikimedia.org/T300442, to keep this conversation focused on changing the Hive backend.

nshahquinn:
The code in the pull request is nice and ready to merge, but won't work because [T292699](https://phabricator.wikimedia.org/T292699) prevents `sasl` from installing correctly. We're blocked on that.

ottomata:
Alright, can you try again?  I just deployed a new version of anaconda-wmf with fixes to the scrips, as well as making cyrus-sasl and pyhive[hive] explicit dependencies in anaconda-wmf.

nshahquinn:
Okay, I tested this again on stat1006. I was able to install it and query successfully both in a new environment and in one I had created before @ottomata rolled out the fix. I also tried querying 3-4 different Hive tables, and it worked as expected.

However, I tried running the test suite (currently merged to master but not this branch) and ran into some problems with `load_csv`. I'll comment in line about those.

wmfdata/hive.py
```diff
@@ -170,34 +149,40 @@ def load_csv(
...
         if headers:
-            __, tmp_path = tempfile.mkstemp()
             with open(path, 'r') as source, open(tmp_path, 'w') as target:
                 # Consume the first line so it doesn't make it to the copy
                 source.readline()
                 copyfileobj(source, target)
             path = tmp_path
 
+        # Copy the file to HDFS because pyhive runs via JDBC not off the local client
+        hdfs_path = f"/tmp/{path}"
```
nshahquinn: There are a couple issues here.

1. `msktemp` returns a path like `/tmp/tmpdw4b3ubr`. That would give us `/tmp//tmp/tmpdw4b3ubr` as our `hdfs_path`.
2. When I tried running this, I got an error because `hdfs:///tmp/tmp` is owned by milimetric, probably because it was created with default permissions when testing this code. If we put the file directly in `/tmp` (we can even use the `mkstemp` path without modification for HDFS), that should avoid the problem.

milimetric: good point, I just removed the '/tmp' prefix I think that fixes it.

milimetric: @nshahquinn I tested this on my notebooks and it seems like it's working ok.  Let me know if there's anything else.

ottomata:
Great, so this should be ready to merge, ya?

milimetric:
Yes, shall we officially do the "Andrew and Dan are maintainers and can
merge" thing now?

nshahquinn:
I just re-ran the test suite and it passed this time, so, for what it's worth, I agree—ready to merge!

> Yes, shall we officially do the "Andrew and Dan are maintainers and can merge" thing now?

Yes, please do! I'd suggest "squash and merge" (🎃&↗↑) in this case.

nshahquinn (approved):

PR #25, opened 2021-12-16
Support loading a Parquet file into a Hive table

nshahquinn:


wmfdata/hdfs.py
```diff
@@ -0,0 +1,29 @@
...
+        check=True
+    )
+
+def delete(hdfs_path):
+    """
+    Deletes the single file in our Hadoop file system located at `hdfs_path`.
+    `hdfs_path` does not need to include the scheme ("hdfs://"); relative paths
+    are interpreted relative to your user directory.
+
+    Deleting directories or multiple files is not supported.
```
milimetric, line 23: I haven't tried but can't you just pass a wildcard in the path and delete multiple files this way?  Maybe it could "ls" the path and count the results to make sure?

wmfdata/spark.py
```diff
@@ -331,10 +343,47 @@ def run(commands, format="pandas", session_type="yarn-regular", extra_settings={
...
+    Loads a Parquet file into the Data Lake and registers it as a Hive table.
+
+    Arguments:
+    * `local_path`: the path (relative or absolute) to the file you want to load
+    * `hive_db`: the Hive database into which to place the loaded data. Must
+      already exist.
+    * `hive_table` (optional): the Hive table name to use for the loaded data.
+      If not given, the local file name is used (e.g. "experiment_data"
+      for "/dir/experiment_data.parquet").
+    * `overwrite` (optional): whether to overwrite the specified Hive table
```
milimetric, line 313: I was thinking if this is specified, the table name should be mandatory too, otherwise you might more easily accidentally overwrite an existing table.  It could still happen but maybe it makes you double check.

wmfdata/spark.py
```diff
@@ -331,10 +343,47 @@ def run(commands, format="pandas", session_type="yarn-regular", extra_settings={
...
+
+    df_writer = df.write
+    # By default, the writer raises an error if the table already exists
+    if overwrite:
+        df_writer.mode("overwrite")
+    df_writer.saveAsTable(f"{hive_db}.{hive_table}")
+
+    # Since we created a "managed" Hive table, Hive has copied the data to its
+    # own HDFS directory. We can delete our temporary copy.
+    hdfs.delete(hdfs_path)
```
milimetric, line 340: Hm, this approach means we write the file to hdfs twice.  If we were just putting it to hdfs and making a hive table on top of it we'd avoid that, is that not possible if you don't know the schema?

wmfdata_tests/tests.py
```diff
@@ -51,9 +51,7 @@ def assert_dataframes_match(df1, df2):
     assert df1.columns.equals(df2.columns)
 
 def set_up_table_1(): 
-    spark = wmf.spark.get_session(type="local", app_name="wmfdata-test")
-    spark_df = spark.read.load(f"file://{this_directory}/test_data_1.parquet")
-    spark_df.write.mode("overwrite").saveAsTable(f"{hive_db}.wmfdata_test_1")
```
milimetric, line 52: Oh, this is what I was thinking would be better, no?

wmfdata/hdfs.py
```diff
@@ -0,0 +1,29 @@
...
+
+def put(local_path, hdfs_path):
+    """
+    Puts the local file located at `local_path` into our Hadoop file system
+    at `hdfs_path`. `hdfs_path` does not need to include the scheme ("hdfs://");
+    relative paths are interpreted relative to your user directory.
+    """
+    check_kerberos_auth()
+    subprocess.run(
+        ["hdfs", "dfs", "-put", local_path, hdfs_path],
```
ottomata, line 13: Not for this change, but FYI we are starting to use the [pyarrow HDFS](https://arrow.apache.org/docs/python/filesystems.html#filesystem-hdfs) interface everywhere.  Even better is [fsspec's pyarrow wrapper](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/arrow.py#L190), allowing to use a pythonic file interface for HDFS, but also for almost any 'filesystem' like URL, including HDFS>

ottomata: Oh, wait, hdfs.py is new?  Maybe you should use fsspec or pyarrow for this change.

wmfdata/spark.py
ottomata: It is possible, but as long as the table is small, this might be preferred.  Doing it this way makes the table a hive managed table (vs an external one), so e.g. DROPing the table will actually delete the data.

wmfdata/spark.py
```diff
@@ -331,10 +343,47 @@ def run(commands, format="pandas", session_type="yarn-regular", extra_settings={
...
+    random_string = uuid.uuid4()
+    hdfs_path = f"hdfs:///tmp/{random_string}-{hive_table}"
+    hdfs.put(local_path.resolve(), hdfs_path)
+    df = session.read.parquet(hdfs_path)
+
+    df_writer = df.write
+    # By default, the writer raises an error if the table already exists
+    if overwrite:
+        df_writer.mode("overwrite")
+    df_writer.saveAsTable(f"{hive_db}.{hive_table}")
```
ottomata, line 336: I don't recall what df_writer will use as the default format.  I guess if it it starts as parquet we probably should write as parquet too?  So maybe add `.format('parquet')`?

wmfdata/spark.py
```diff
@@ -331,10 +343,47 @@ def run(commands, format="pandas", session_type="yarn-regular", extra_settings={
...
+        start_session_timeout(spark_session)
 
     return overall_result
+
+def load_parquet(local_path, hive_db, hive_table=None, overwrite=False):
+    """
+    Loads a Parquet file into the Data Lake and registers it as a Hive table.
+
+    Arguments:
+    * `local_path`: the path (relative or absolute) to the file you want to load
```
ottomata, line 307: Why support only local paths?  If you use fsspec, this could be any fully qualified URL as long as it has an fsspec implementation.  

Here's an [example](https://gitlab.wikimedia.org/repos/data-engineering/workflow_utils/-/blob/main/workflow_utils/artifact/cache.py#L113-115) of how I'm using it to sync artifacts to hdfs from almost anywhere.  Basically:

```python
with fspec.open(source_url).open as input_stream:
    with fsspec.open(dest_url, mode='wb').open() as output_stream:
        output_stream.write(input_stream.read())
```

in this case input_url could be anything, and output_url can be your `hdfs:///` (or probably better `arrow_hdfs:///` path. (ATM, fsspec will use a deprecated pyarrow API with just 'hdfs:///' URLs.  Marcel and I just realized this and will figure out how best to use fsspec + pyarrow + hdfs soon).

Anyway, I'm not sure if I am recommending for you to do this, but just wanted you to be aware that it is possible.
Likely you'd only ever have parquet files locally or already in HDFS though, so you could just vary how to read and then put the file (I guess i the file is already in HDFS you don't need to put it there) based on URI proto.

nshahquinn:
I just rebased this on `main`. Still not sure when I'll have a chance to finish this, but it's not completely dead 😅

PR #26, opened 2022-01-29
Ensure binary data in MariaDB is returned as strings

nshahquinn:
Previously, we relied on the fact that mysql-connector-python
would convert any binary data to strings whenever the connection
was set to use the UTF-8 character set. This behavior was changed
in version 8.0.24, so that binary data is returned as bytearrays
regardless of the specified character set (although the setting
still causes field names to be converted).

This change adds a converter class that explicitly converts any
binary data to strings.

### DIFF
--- a/wmfdata/mariadb.py
+++ b/wmfdata/mariadb.py
@@ -12,6 +12,20 @@ import pandas as pd
 
 from wmfdata.utils import ensure_list
 
+# https://stackoverflow.com/a/68784172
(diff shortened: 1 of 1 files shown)


nettrom:
Tested querying user names and page titles in various wikis and found no issues, looks good to me!

nettrom (approved):
Tested this querying user names and page titles in various wikis. Found no issues, looks good to me!

PR #27, opened 2022-01-29
Add missing argument to session timeout

nshahquinn:
As far as I know, the timeouts never functioned properly because of this missing argument.


nshahquinn:
> As far as I know, the timeouts never functioned properly because of this missing argument.

Ah, no, it looks like the problem was introduced in 80dd085, which got released in version 1.1.

nettrom (approved):
Tested and verified that it doesn't result in any error messages and that the Spark session is stopped after the defined timeout. Looks good to me!

PR #28, opened 2022-02-02
Fix decoding of binary text from MariaDB

nshahquinn:
My previous approach resulted in all fields (including numeric ones) being returned as strings, because MySQL Connector Python internally handles all fields as bytearrays. This patch instead looks up the actual data type of the field and calls the decode function only if it's actually a string field. This code doesn't convert [blob fields](https://mariadb.com/kb/en/blob-and-text-data-types/), but it barely matters because I believe MediaWiki doesn't use them at all.

### DIFF
--- a/wmfdata/mariadb.py
+++ b/wmfdata/mariadb.py
@@ -15,16 +15,16 @@ from wmfdata.utils import ensure_list
 # https://stackoverflow.com/a/68784172
 class BytesConverter(mysql.conversion.MySQLConverter):
     """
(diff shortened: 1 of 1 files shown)


nshahquinn:
Unlike last time, I ran the test suite _before_ putting up this request, so I can confirm it passes 😅

nettrom (approved):
Verified that strings (e.g. user names, page titles) are correctly converted, and that numbers (e.g. page and user IDs) are kept as numbers.

PR #29, opened 2022-02-07
Improve handling of nulls and blobs in binary fields

nshahquinn:
Previously, a null value in a binary-encoded MariaDB field would cause an
AttributeError. These are rare in our infrastructure since MediaWiki generally
uses empty strings instead.

I had assumed that MediaWiki did not use blob fields, but it turns out they
are very common. In all the cases I found, the fields contain UTF-8 string
data, so these are now decoded as well.

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -3,8 +3,9 @@ This document has useful instructions for people who are interested fixing bugs 
 ## Installing in development mode
 When you work on changes to this package, you will want an easily-accessible local copy that's at the top of your Python path. Here's the recommended way to do that:
 1. Clone this repository to whatever location you like.
-1. Install the cloned repository in [development mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) using `pip install -e`.
+1. Install the cloned repository in [development mode](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) using `pip install -e .`
     * This will end early with the message `ERROR: Could not install packages due to an OSError: [Errno 30] Read-only file system: 'WHEEL'`. This is okay; the crucial step has already been done.
+    * NOTE: In some cases, this can also give you `ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied...`.  In this case, you *should* follow the advice and pass `--user` as in `pip install --user -e .`
 1. Change to the `site-packages` directory for your current environment using the following command: `` cd `python -c "import sysconfig; print(sysconfig.get_path('purelib'))"` ``
 1. Edit the `anaconda.pth` file in that directory to include the absolute path (not using `~`) to the cloned repository.
     * This ensures the new version of Wmfdata is on your Python path (which `pip install -e` would have done if not for the error) _and_ ensures it takes precendence over the version of Wmfdata in the base `anaconda-wmf` environment.
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(
         "IPython",
         "findspark",
         "matplotlib>=2.1", # 2.1 introduced ticker.PercentFormatter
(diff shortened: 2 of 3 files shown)


nshahquinn:
Just verified that this passes the test suite.

nshahquinn:
After I noticed the null value problem, I went diving through lots of different MariaDB tables looking for other cases that weren't being handled properly. That's how I realized that there were actually lots of blob fields. So, _hopefully_, now that I've gone looking, I've found all the substantial issues in this area.

nettrom (changes requested):
I don't get this to pass testing. From what I can tell, any integer column (e.g. `content_edits` in `wmfdata_test`, `user_editcount` in `user`) that has a value of `0` results in a `NaN`, and so the MariaDB equality test fails (on `content_edits` rows 6 and 10 in `wmfdata_test`).

It look like this happens because the call in line 37 to `super().to_python()` results in a match on the if-statement in https://github.com/mysql/mysql-connector-python/blob/master/lib/mysql/connector/conversion.py#L190, and so a `None` is returned from that method. If I comment out line 94 so `ByteConverter` isn't used, zero-value integers are returned correctly.

Passing conversion up to the superclass seems like the intuitive thing to do. I'm not sure if the superclass expects us to identify the returned column types and call their respective methods rather than expect `to_python()` to behave in a more reasonable way.

nshahquinn:
> I don't get this to pass testing. From what I can tell, any integer column (e.g. `content_edits` in `wmfdata_test`, `user_editcount` in `user`) that has a value of `0` results in a `NaN`, and so the MariaDB equality test fails (on `content_edits` rows 6 and 10 in `wmfdata_test`).
>
> It look like this happens because the call in line 37 to `super().to_python()` results in a match on the if-statement in https://github.com/mysql/mysql-connector-python/blob/master/lib/mysql/connector/conversion.py#L190, and so a `None` is returned from that method. If I comment out line 94 so `ByteConverter` isn't used, zero-value integers are returned correctly.

Okay, wow! This is confusing. I didn't see this with my installation, with mysql-connector-python 8.0.28. However, I replicated what you saw after switching to a brand-new environment, with mysql-connector-python 8.0.23.

It's not immediately clear why the behavior is so different; both version have the line you identified as the culprit. The only change to `conversion.py` between the two versions was the [one which set off this string of MariaDB patches](https://github.com/mysql/mysql-connector-python/commit/82764da6cb062eb890e311049e700315a190a9f8), so it's almost certainly responsible. I spent almost an hour trying to figure out exactly how, but I gave up after reflecting that it doesn't matter since it's easy enough to require the newer version of mysql-connector-python.

I'll push that change up now; hopefully it works for you!

nshahquinn:
Okay, I hadn't actually pushed the change (maybe I should add commits to the branch instead of amending one commit like I would on Gerrit). But _now_ I have.

nshahquinn:
Thank you very much @milimetric!

It's not a big deal at all, but when you have an unrelated change to make, I suggest making a separate PR or commit. In this case, since it was a documentation improvement that didn't touch code, I think you can just make a separate commit directly to the master branch, without making a PR and having it review 😊

nshahquinn:
> It's not a big deal at all, but when you have an unrelated change to make, I suggest making a separate PR or commit. In this case, since it was a documentation improvement that didn't touch code, I think you can just make a separate commit directly to the master branch, without making a PR and having it review 😊

Oh, I see from your messages on Slack that it's related to getting the right version of mysql-connector-python installed. That makes sense, then. Ignore me 😊

milimetric:
I would've honestly put it in a different PR but for some reason the way I
did it forced it into this one.  Github is a magical place.

PR #31, opened 2022-10-07
Add function to format an SQL tuple

nshahquinn:
This is a useful convenience for analysts. For example, if you have a long list of users and you want to look up their user IDs, this lets you simply write:
```python
wmf.mariadb.run(f"""
SELECT
    user_name,
    user_id
FROM user
WHERE
    user_name IN {sql_tuple(users)}
""", "enwiki")
```

### DIFF
--- a/wmfdata/utils.py
+++ b/wmfdata/utils.py
@@ -175,19 +175,30 @@ def sql_tuple(i):
     > sql_tuple(["a", "b", "c"])
     "('a', 'b', 'c')"
     """
-    # It might seem useful to return "()" when an empty iterable is passed, but "IN ()"
-    # results in an SQL syntax error. Instead, we should send a clear signal to the caller
-    # that it needs to better handle the no-items case.
-    if len(i) == 0:
-        raise ValueError("Cannot produce an SQL tuple without any items.")
 
     # Transform other iterables into lists, raising errors for non-iterables
     if type(i) != list:
         i = [x for x in i]
 
+    # It might seem useful to return "()" when an empty iterable is passed, but "IN ()"
+    # results in an SQL syntax error. Instead, we should send a clear signal to the caller
+    # that it needs to better handle the no-items case.
+    #
+    # Running this check after converting i to a list avoids errors from generators that
+    # don't implement len.
+    if len(i) == 0:
+        raise ValueError("Cannot produce an SQL tuple without any items.")
+
     # Using Python's string representation functionality means we get escaping for free. Using the
     # representation of a tuple almost works, but fails when there's just one element, because SQL
     # doesn't accept the trailing comma that Python uses. Instead, we use representation of a
     # list and replace the brackets with parentheses.
+    #
+    # To-do: The string representation method fails in an edge case. It normally produces single
+    # quotes but falls back to double quotes if the string contains a single quote. If the database
+    # strictly follows the ANSI SQL standard (as Presto does), double-quoted strings are
+    # interpreted as literal column names rather than strings. This is low priority, but ideally
+    # we would fix this by constructing the SQL tuple manually rather than using the string
+    # representation.
     list_repr = repr(i)
     return "(" + list_repr[1:-1] + ")"
--- a/wmfdata/utils.py
+++ b/wmfdata/utils.py
@@ -166,4 +166,28 @@ def python_version():
     """
     return f"{sys.version_info.major}.{sys.version_info.minor}"
 
+def sql_tuple(i):
+    """
+    Given a Python iterable, returns a string representation that can be used in an SQL IN
+    clause.
 
+    For example:
+    > sql_tuple(["a", "b", "c"])
+    "('a', 'b', 'c')"
+    """
+    # It might seem useful to return "()" when an empty iterable is passed, but "IN ()"
+    # results in an SQL syntax error. Instead, we should send a clear signal to the caller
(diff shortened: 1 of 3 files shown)


xabriel:
Great stuff. Should we add a unit test?

nshahquinn:
> Great stuff. Should we add a unit test?

That would definitely be proper, but so far we haven't set up any CI or unit testing framework. (There is a smoke test suite in `/wmfdata-tests` that is run manually on one of the analytics clients.) So it seems like a much bigger project than just adding a single unit test.

I would suggest merging this as-is and saving that for later (after we move to GitLab with its new CI system, for one thing: [T304544](https://phabricator.wikimedia.org/T304544)), unless there's a quick version we can do now (maybe you already have a framework in mind, and the tests can just be run manually for now?)

xabriel:
> unless there's a quick version we can do now

Agreed that adding a unit test framework is out of scope for this PR. How about just adding a test to `wmfdata_tests/tests.py` and running it manually?

nshahquinn:
> > unless there's a quick version we can do now
> 
> Agreed that adding a unit test framework is out of scope for this PR. How about just adding a test to `wmfdata_tests/tests.py` and running it manually?

Hah, yeah! Why didn't I think of that? I'll put that up shortly.

nshahquinn:
Well, apart from the test itself, writing the test made me notice a couple edge cases I hadn't considered. Maybe unit testing is a good idea? Imagine that! 😊

I'm not very experienced with testing, so please point out anything important I've missed. This whole testing setup is definitely very hacky and switching to an actual framework would be a good thing, but that's a challenge for another time.

wmfdata/utils.py
```diff
@@ -175,19 +175,30 @@ def sql_tuple(i):
...
     # list and replace the brackets with parentheses.
+    #
+    # To-do: The string representation method fails in an edge case. It normally produces single
+    # quotes but falls back to double quotes if the string contains a single quote. If the database
+    # strictly follows the ANSI SQL standard (as Presto does), double-quoted strings are
+    # interpreted as literal column names rather than strings. This is low priority, but ideally
+    # we would fix this by constructing the SQL tuple manually rather than using the string
+    # representation.
     list_repr = repr(i)
     return "(" + list_repr[1:-1] + ")"
```
xabriel, line 204: You could avoid the edge case by doing instead:

```
>>> iter = ['enwiki', 'eswiki', 'ptwiki']
>>> quoted_iter = [f"'{e}'" for e in iter]
>>> '(' + ', '.join(quoted_iter) + ')'
"('enwiki', 'eswiki', 'ptwiki')"
```

wmfdata_tests/tests.py
```diff
@@ -171,6 +171,20 @@ def test_load_csv():
...
+    try:
+        wmf.utils.sql_tuple([])
+    # We want a ValueError in this case
+    except ValueError:
+        pass
+    else:
+        raise AssertionError("Passing an empty iterable to sql_tuple should raise a ValueError.")
+
+    t = wmf.utils.sql_tuple(("enwiki", "arwiki", "dewiki"))
+    assert t == "('enwiki', 'arwiki', 'dewiki')"
```
xabriel, line 184: You covered empty case and multiple elements case, which is great. It would have also been nice to cover the single element case (which, as you had alluded elsewhere, would expose your non-ANSI SQL edge case).

xabriel (commented):
This one is almost ready, thank you for all the changes @nshahquinn!

I just want you to consider my comments above, but I am ready to merge as is if you're ok with it. LMK.

wmfdata/utils.py
nshahquinn: Unfortunately, that doesn't work! One component of the edge case is that the string contains a single quote (`'`). With this, you're wrapping in single quotes, so another single quote in the middle will break it early and cause problems.

Since you nerd-sniped me (😁), I started looking more deeply into what needs to be escaped, and it turns out the problem is even bigger than I realized: there's no single escaping strategy that works for all of MariaDB, Spark, and Presto, so for this to work correctly in all cases, we'd have to implement more than one (ourselves or with some library) and let the user choose. More details: https://github.com/nshahquinn/misc-wikimedia-analysis/blob/master/2022-10_SQL_string_escaping.ipynb

Much as I want to, I can't justify spending more time on this now, so I think the best strategy is just to stick with the current implementation. As far as I can tell, it always works as long no single quotes or backlashes get involved, and even then it works about half the time. I will update the docstring and comments with what I've learned, though.

wmfdata_tests/tests.py
nshahquinn: Well, the main component of the edge cases (since I now noticed another one) is not having a single element but having an element than includes a single quote or a backslash. I could add tests for those but they would always fail, so that doesn't seem very helpful. Is there something I'm missing?

wmfdata/utils.py
xabriel: > I think the best strategy is just to stick with the current implementation

Agreed. Definitely a fun problem... that we can solve later!

wmfdata_tests/tests.py
xabriel: > I could add tests for those but they would always fail

Actually, failing tests are useful too. They document limitations of the implementation. But we can get into those once we add a unit test suite.

xabriel:
Merging this!

PR #30, opened 2022-10-06
Switch from MySQL to MariaDB connector library

nshahquinn:
I tested this with version 1.0.11 of the MariaDB connector library, which is the latest which works with the version of the MariaDB C library that's installed on the analytics clients (3.1.16). I'm not sure if that library is easily upgradable; if it is, it might be worth doing that and then bumping the version of the Python connector.

PR #32, opened 2022-10-07
Switch from deprecated Hive run function

nshahquinn:
`run_cli` is deprecated and replaced by `run`, so our test suite should use the latter. I've successfully run the test suite with the change (although that required rebasing on the work in #30 to avoid the MariaDB errors).


nshahquinn:
No need to include this in the changelog since it's so minor.

xabriel:
Is this ready to merge @nshahquinn ?

nshahquinn:
> Is this ready to merge @nshahquinn ?

@xabriel yes, I think so!

PR #35, opened 2022-11-07
Remove deprecated and unneeded features

nshahquinn:
* the entire charting module (not formally deprecated)

In the hive module:
* run_cli function
* "raw" results from the run function
* heap_size and engine parameters of the run function
* format parameter of the run function (not formally deprecated)

In the mariadb module:
* "raw" results from the run function
* format parameter of the run function (not formally deprecated)

In the spark module:
* "raw" results from the run function
* format parameter of the run function (not formally deprecated)

Bug: T300442


nshahquinn:
I realized I still need to write the change notes. I'll work on that on Thursday.

wmfdata/spark.py
```diff
@@ -326,10 +315,7 @@ def run(commands, format="pandas", session_type="yarn-regular", extra_settings={
             overall_result = cmd_result
     
     if overall_result:
-        if format == "pandas":
-            overall_result = overall_result.toPandas()
-        elif format == "raw":
-            overall_result = overall_result.collect()
+        overall_result = overall_result.toPandas()
```
xabriel, line 328: `toPandas()` materializes the result set in the driver, and if the driver doesn't have sufficient memory it will OOM. Don't we want an option that just keeps the Spark `DataFrame`? Or is that not the intended use of the `run` method?

In either case it would be good to document the materialization behavior on the method docs.

Also, if we keep just `toPandas()`, should we define the return type?

xabriel (commented):
Thanks for the quick turnaround @nshahquinn !

One comment for your consideration below before we merge.

wmfdata/spark.py
nshahquinn: So the `run` functions here and in the MariaDB, Hive, and Presto modules are designed to be simple "write SQL, get a Pandas DataFrame" interfaces. With any of those you could pull down too much data and get an OOM (just as you could try loading a huge Parquet file into Pandas and get an OOM). So in that sense this function is just behaving as it has always done. I will mention it in the docs, though.

The point about returning a Spark dataframe is interesting! I don't think we need to do it here, since the function has never supported it before and we have enough stuff to take care of in this release 😊 But in the future we could potentially add a "spark" format that does what you suggest. I'm not sure how valuable it would be, since if you are interacting with the PySpark API directly you know how to use its `sql` method. But I'm open to the discussion.

> Also, if we keep just toPandas(), should we define the return type?

Sure, seems like a good idea!

xabriel: 👍

nshahquinn: FYI, I'm going to update the docs and add the type hints in #36, since it seems like it'll be more on topic there.

nshahquinn:
@xabriel okay, I think this is ready to merge. I've rebased it on main, added the entries to the release notes, and ran the test suite.

PR #34, opened 2022-11-03
Default spark sessions to Spark3.

xabriel:
Simple changes that:
* Make Spark3 be the new default for Spark Sessions.
* Explains how to override back to Spark2.

PR #36, opened 2022-11-08
Spark: Redesign API and remove session timeouts

nshahquinn:
* **Breaking change**: the `get_session` and `get_custom_session` functions have been renamed to `create_session` and `create_custom_sesson`. They will now stop any existing session before creating the new session. This means that the returned session will always reflect the passed settings; previously, the settings were silently ignored if a session already existed. Use the new `get_active_session` function if you want to non-destructively retrieve the active session. 
* **Breaking change**: the deprecated "raw" format and the non-deprecated `format` parameter have been removed from the `run` function.
* **Breaking change**: the `run` function no longer has the ability to specify Spark settings, as the `session_type` and `extra_settings` parameters have been removed. If a session already exists, it will be used. Otherwise, a default "yarn-regular" session will be created and used.
* Previously, in some cases, the package automatically closed sessions after 30 minutes of apparent inactivity. It no longer does this.
  * Breaking change: the `get_application_id`, `cancel_session_timeout`, `stop_session`, and `start_session_timeout` functions have been removed. These were just intended for internal use, so it's unlikely that you will have to change code.


xabriel:
> However, I haven't had a chance to test it yet, because I based it on main, which no longer works with the existing Conda setup.

I'll revert #34 then while we work on these changes. These are changes we want to test throughly.

xabriel:
Wait, no need to revert, you should be able to do:
```
import os
os.environ['SPARK_HOME'] = "/usr/lib/spark2"
```

nshahquinn:
@xabriel I've now done everything I intended, so this is ready for your review! The test suite passes.

xabriel:
There is an issue when creating spark sessions back to back that ship the user conda environment:

```
import wmfdata

spark = wmfdata.spark.create_session(
    type='yarn-large',
    ship_python_env=True
)

spark2 = wmfdata.spark.create_session(
    type='yarn-large',
    ship_python_env=True
)
```
creation of `spark2` fails with:
```
A conda environment is already packed at conda-2022-11-14-test-spark-api-changes.tgz. If you have recently installed new packages into your conda env, set force=True in conda_pack_kwargs and it will be repacked for you.
Will ship conda-2022-11-14-test-spark-api-changes.tgz to remote Spark executors.
22/11/15 02:19:03 WARN Utils: Service 'sparkDriver' could not bind on port 12000. Attempting port 12001.
22/11/15 02:19:03 WARN Utils: Service 'SparkUI' could not bind on port 4040. Attempting port 4041.
22/11/15 02:19:09 WARN Client: Same path resource file:/home/xcollazo/notebooks/conda-2022-11-14-test-spark-api-changes.tgz#conda-2022-11-14-test-spark-api-changes added multiple times to distributed cache.
22/11/15 02:19:09 ERROR SparkContext: Error initializing SparkContext.
java.lang.IllegalArgumentException: Attempt to add (conda-2022-11-14-test-spark-api-changes.tgz#conda-2022-11-14-test-spark-api-changes) multiple times to the distributed cache.
	at org.apache.spark.deploy.yarn.Client.$anonfun$prepareLocalResources$23(Client.scala:631)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
	at org.apache.spark.deploy.yarn.Client.$anonfun$prepareLocalResources$22(Client.scala:621)
	at org.apache.spark.deploy.yarn.Client.$anonfun$prepareLocalResources$22$adapted(Client.scala:620)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at org.apache.spark.deploy.yarn.Client.prepareLocalResources(Client.scala:620)
	at org.apache.spark.deploy.yarn.Client.createContainerLaunchContext(Client.scala:887)
	at org.apache.spark.deploy.yarn.Client.submitApplication(Client.scala:202)
	at org.apache.spark.scheduler.cluster.YarnClientSchedulerBackend.start(YarnClientSchedulerBackend.scala:62)
	at org.apache.spark.scheduler.TaskSchedulerImpl.start(TaskSchedulerImpl.scala:220)
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:579)
	at org.apache.spark.api.java.JavaSparkContext.<init>(JavaSparkContext.scala:58)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:247)
	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
	at py4j.Gateway.invoke(Gateway.java:238)
	at py4j.commands.ConstructorCommand.invokeConstructor(ConstructorCommand.java:80)
	at py4j.commands.ConstructorCommand.execute(ConstructorCommand.java:69)
	at py4j.GatewayConnection.run(GatewayConnection.java:238)
	at java.lang.Thread.run(Thread.java:750)
22/11/15 02:19:09 WARN YarnSchedulerBackend$YarnSchedulerEndpoint: Attempted to request executors before the AM has registered!
22/11/15 02:19:09 WARN MetricsSystem: Stopping a MetricsSystem that is not running
```

It seems like, for some reason, Spark is caching the `spark.yarn.dist.archives` config.

xabriel:
I have a patch for the issue above, will share in a bit.

wmfdata/spark.py
```diff
@@ -129,20 +129,27 @@ def create_custom_session(
             category=FutureWarning
         )
 
+    old_session = get_active_session()
+    if old_session:
+        old_session.stop()
```
nshahquinn: I notice you switched from stopping the context to stopping the session. Just curious: did you do this just because they're basically the same thing and this lets you reuse `get_active session`? Or is there some relevant difference here?

xabriel:
All right, on 42d0200 I fixed the issue from https://github.com/wikimedia/wmfdata-python/pull/36#issuecomment-1314671723, and on 699c32e I added some sanity tests for these new APIs.

I also tested this PR on both an existing `anaconda-wmf` environment as well as on the latest `conda-analytics`, and found no further issues.

xabriel:
@ottomata, could you give this PR a check?

xabriel (approved):
Looks good to me, but would like @ottomata's review as well.

wmfdata/spark.py
```diff
@@ -129,20 +129,27 @@ def create_custom_session(
...
             # The path to our packed conda environment.
             conda_packed_file = conda.pack(**conda_pack_kwargs)
             # This will be used as the unpacked directory name in the YARN working directory.
             conda_packed_name = os.path.splitext(os.path.basename(conda_packed_file))[0]
 
-            # Ship conda_packed_file to each YARN worker.
+            # Ship conda_packed_file to each YARN worker, unless a previous session has done it already.
             conda_spark_archive = f"{conda_packed_file}#{conda_packed_name}"
-            if "spark.yarn.dist.archives" in spark_config:
+            if not old_session and "spark.yarn.dist.archives" in spark_config:
```
nshahquinn: There will be some cases where the user has made changes to the environment and will want to reship it. Currently, it looks like there's no way to do this without manually running `conda.pack({"force": True})`; even restarting your Jupyter kernel doesn't remove the cached environment.

Should we force a repack every time? It wouldn't be too hard to do that by altering`conda.pack` to add a random string to the filename so the output is different every time. But on the other hand that adds scope and repacking can be a pretty time-consuming operation.

Maybe we should add an argument to `create_custom_session` that forces a repack? Hmm.

wmfdata_tests/tests.py
```diff
@@ -119,6 +119,18 @@ def test_silent_command_via_hive():
    assert output is None
    log_test_passed("Silent command via Hive")
 
+def test_spark_session_apis():
```
nshahquinn, line 122: Thank you! This is great 😊

wmfdata/spark.py
xabriel: > did you do this just because they're basically the same thing and this lets you reuse get_active session?

This.

wmfdata/spark.py
xabriel: > Currently, it looks like there's no way to do this without manually running conda.pack({"force": True})

You can do this today:
```python
create_custom_session(
    ...,
    ship_python_env=True,
    conda_pack_kwargs={"force": True}
)
```
But yes you have to use `create_custom_session()` instead of `create_session()`, which I think is reasonable since `create_session()` is a simplified API.

Am I missing something?

wmfdata/spark.py
```diff
@@ -12,20 +13,20 @@
...
 SPARK_HOME = os.environ.get("SPARK_HOME", default_spark)
-if SPARK_HOME == "/usr/lib/spark2":
-    warnings.warn(
-        "Spark2 has been deprecated. Please upgrade your jobs to Spark3. "
-        "See https://wikitech.wikimedia.org/wiki/Analytics/Systems/Cluster/Spark/Migration_to_Spark_3 for details.",
-        category=FutureWarning
-    )
+
+# This is not necessary in a conda-analytics environment. Once we stop supporting anaconda-wmf
+# environments, we can drop this line and the dependency on findspark.
```
ottomata, line 24: Great, I think that would be best!  If findspark is needed, the user can call it in their script ahead of time.

wmfdata/spark.py
ottomata: Just a guess: The Spark (SQL) Session is a wrapper around the SparkContext, so stopping the session also stops the SparkContext, and is probably the better thing to do because we mostly interact with the SparkSession anyway.

wmfdata/spark.py
```diff
@@ -74,22 +75,30 @@
...
+def get_active_session():
+    """
+    Returns the active session if there is one and None otherwise.
+
+    Spark 3 includes a native getActiveSession function. Once we end Spark 2
+    support, we can switch to that.
+    """
+    if SparkContext._jvm:
+        if SparkContext._jvm.SparkSession.active():
+            return SparkSession.builder.getOrCreate()
```
ottomata, line 88: return None explicitly? if not active?

wmfdata/spark.py
ottomata: Why do we need to check `old_session` at all here?  You've  stopped the old one above, so you'll need to do this either way to create the new session, no?

wmfdata/spark.py
```diff
@@ -143,26 +167,9 @@ def get_custom_session(
         elif os.path.isfile(os.path.join(f"/usr/bin/python{python_version()}")):
             os.environ["PYSPARK_PYTHON"] = f"/usr/bin/python{python_version()}"
 
-        if "PYSPARK_PYTHON" in os.environ:
```
ottomata, line 146: Why remove this message?  Things can get pretty weird if an unexpected python is used.

wmfdata/spark.py
```diff
@@ -112,20 +121,35 @@ def get_custom_session(
...
             conda_packed_name = os.path.splitext(os.path.basename(conda_packed_file))[0]
 
-            # Ship conda_packed_file to each YARN worker.
+            # Ship conda_packed_file to each YARN worker, unless a previous session has done it already.
             conda_spark_archive = f"{conda_packed_file}#{conda_packed_name}"
-            if "spark.yarn.dist.archives" in spark_config:
+            if not old_session and "spark.yarn.dist.archives" in spark_config:
                 spark_config["spark.yarn.dist.archives"] += f",{conda_spark_archive}"
-            else:
+            elif not old_session:
```
ottomata: (If we do need to check `old_session`), you are currently are checking `not old_session` twice here. I'd find this more readable like:

```python
if not old_session:
    if "spark.yarn.dist.archives" in spark_config:
        # ...
    else:
        # ....
```

wmfdata/spark.py
xabriel: Ah wait, yes, this new comment is misleading:
https://github.com/wikimedia/wmfdata-python/blob/699c32e3322e9de38d7ff841a5fba5f731d97ccc/wmfdata/spark.py#L150-L152

What is really happing is that I am not modifying the list at `spark.yarn.dist.archives`, but the archive will be uploaded every time a new Yarn App is spawned, which happens in tandem with creating a new SparkContext. See Yarn Client code [here](https://github.com/apache/spark/blob/7955b3962ac46b89564e0613db7bea98a1478bf2/resource-managers/yarn/src/main/scala/org/apache/spark/deploy/yarn/Client.scala#L635). 

Will fix comment.

xabriel: > Why do we need to check old_session at all here?

Because we found this weird behavior where the `SparkContext` is gone, but its `SparkConf` seems to persist. So if we set `spark.yarn.dist.archives` previously, we want to avoid setting it again to not get the stack at https://github.com/wikimedia/wmfdata-python/pull/36#issuecomment-1314671723.

wmfdata/spark.py
nshahquinn: PySpark prints essentially the same message already when starting a session: for example, `PYSPARK_PYTHON=/usr/lib/anaconda-wmf/bin/python3`.

nshahquinn: Although, weirdly, it only prints it the first you create a session in a notebook. Subsequent times the line gets omitted, even if the `PYSPARK_PYTHON` is different.

wmfdata/spark.py
nshahquinn: Okay, makes sense. I like the new approach!

wmfdata/spark.py
nshahquinn: I just fixed `get_active_session` so that it doesn't throw an error when the most recent session was stopped. While I was added, I made the return explicit in all cases.

nshahquinn:
I get a `CondaPackException` when I try to create a session with a shipped `conda-analytics` environment on an-test-client1001. This must be related to upgrading Wmfdata, because I'm pretty sure that's the only thing I did.

Full error: https://phabricator.wikimedia.org/P39868

xabriel:
> I get a `CondaPackException` when I try to create a session with a shipped `conda-analytics` environment on an-test-client1001. This must be related to upgrading Wmfdata, because I'm pretty sure that's the only thing I did.
> 
> Full error: https://phabricator.wikimedia.org/P39868

This is a repro of https://gitlab.wikimedia.org/repos/data-engineering/conda-analytics/-/merge_requests/13. TL;DR: conda-pack 0.6.0 is not happy with Python 3.10. Fix is to bump conda-pack to 0.7.0.

A new `conda-analytics` is now installed on `an-test-client1001` with a fix for this. Please delete your previous `conda-analytics` clones and create a new one.

wmfdata/spark.py
xabriel: Agreed, will fix.

wmfdata/spark.py
xabriel: Fixed misleading `print_err()` on [5daacf3](https://github.com/wikimedia/wmfdata-python/pull/36/commits/5daacf34f65f751035e52f0156b3a6ea510f6009). Also added comment to make the `spark.yarn.dist.archives` hack explicit.

wmfdata/spark.py
xabriel: Fixed.

wmfdata/spark.py
xabriel: > > Why do we need to check old_session at all here?
> 
> Because we found this weird behavior where the `SparkContext` is gone, but its `SparkConf` seems to persist. So if we set `spark.yarn.dist.archives` previously, we want to avoid setting it again to not get the stack at [#36 (comment)](https://github.com/wikimedia/wmfdata-python/pull/36#issuecomment-1314671723).

Further, I found no cleaner way to deal with this as the `spark.yarn.dist.archives` gets set even when I create a `SparkConf` manually. I also tried `SparkConf(loadDefaults=False)`, and that works but then we loose all the good defaults that we have set on our `spark-defaults.conf`.

ottomata: Wow nasty. Hm.  Since we are only trying to work around the problem of SparkConf persisting, and we are only messing with the `spark.yarn.dist.archive`,  instead of checking `old_session`, perhaps we should check if `{conda_spark_archive} not in SparkConf.get("spark.yarn.dist.archives")` ?  That seems to more explicitly address the problem, rather assuming this config is in SparkConf if a session was previously created.

xabriel: Nasty indeed!

Here is the latest version after offline discussion with @ottomata:
https://github.com/wikimedia/wmfdata-python/blob/c95ad645a72d0b85f5b13d7b5a0cd7bfee9145f4/wmfdata/spark.py#L155-L168

ottomata: YES  explicit and very readable, LGTM!

ottomata (approved):

nshahquinn:
Okay, this is looking good! I'm going to give it some final testing now and, if there are no objections, merge it.

xabriel:
> if there are no objections, merge it

LGTM!

wmfdata/spark.py
```diff
@@ -112,20 +131,41 @@ def get_custom_session(
...
+            # Therefore, we should only set 'spark.yarn.dist.archives' if the SparkConf doesn't have it
+            # as otherwise we will get a dist cache exception.
+            # see https://github.com/wikimedia/wmfdata-python/pull/36#discussion_r1023307058
+            is_archive_set = old_session and \
+                old_session.sparkContext.getConf().contains("spark.yarn.dist.archives")
+            if not is_archive_set:
+                if "spark.yarn.dist.archives" in spark_config:
+                    spark_config["spark.yarn.dist.archives"] += f",{conda_spark_archive}"
+                else:
+                    spark_config["spark.yarn.dist.archives"] = conda_spark_archive
```
nshahquinn, line 171: Hmm, you still get the "attempt to add X multiple times to the distributed cache" error if you:
1. Create a session with a shipped environment
2. Manually stop the session
3. Attempt to create a new session with a shipped environment

There's no `old_session` since it was already stopped, so it attempts to ship the packed environment again.

Test case and error: https://phabricator.wikimedia.org/P40122

nshahquinn: Ahh, I think I see a way to fix this. Commit incoming!

xabriel: Isn't this a far fetched failure case? Either we let the new APIs take care of the `SparkSession` lifecycle, or we don't. And if we don't, then we don't need these new APIs. Right?

We could make this explicit on the API docs.

nshahquinn: It won't be that common, but I don't think it's far-fetched. We definitely intend for people to create sessions and then use the session methods directly, and one of those methods is `stop()`. Besides, sessions sometimes die unexpectedly due to errors, and that would leave us in the same state (archive previously shipped but no active session).

xabriel: Fair enough.

I had discussed with @ottomata the possibility of just creating a toss away `SparkConf`. That would be independent of the session. So a potential solution for the current issue.

nshahquinn: > I had discussed with @ottomata the possibility of just creating a toss away SparkConf. That would be independent of the session. So a potential solution for the current issue.

Yeah, I was thinking about that too, but I don't actually see a way to do that. If you call `pyspark.SparkConf()` you still get the old settings rather than  a fresh config. But I think [be853b0](https://github.com/wikimedia/wmfdata-python/pull/36/commits/be853b0b59cc74f30c1ee464a34b11663b1c9afe) solves it—take a look and tell me what you think!

xabriel: Perfect!

xabriel (approved):
LGTM!

xabriel:
I did another pass of tests of this PR against an environment with `anaconda-wmf` and Spark2, and  with `conda-analytics` and Spark3, and all my ad-hoc tests were successful, including the issue discussed in https://github.com/wikimedia/wmfdata-python/pull/36#discussion_r1025722852.

nshahquinn:
Great! Then I'm going to press the big green button now. Thanks for the great collaboration, @xabriel and @ottomata 🎉

PR #37, opened 2022-11-09
Find conda when on a cloned conda-analytics env.

xabriel:
On the new `conda-analytics` conda environment, we do not have the conda package available.

This is because when doing `conda create --clone ...` conda explicitly removes conda related packages. I presume the rationale is that only the base should typically include the `conda` package. This was never an issue on `anaconda-wmf` since that environment was using the `--stacked` flag.

In this PR we thus modify the `sys.path` to find conda on the base environment, if we detect that we are running on top of `conda-analytics`.

### DIFF
--- a/wmfdata/conda.py
+++ b/wmfdata/conda.py
@@ -1,19 +1,32 @@
 import json
 import os
-
-# In case conda is not installed, we still want these functions to return something useful.
-try:
-    from conda.cli.python_api import run_command as condacli
-    conda_installed = True
-except ImportError:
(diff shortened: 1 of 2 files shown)


xabriel:
Alternatively, we could just `conda install conda` on the cloned environment, although that seems to be discouraged? @ottomata WDYT?

xabriel:
Ok discussed this with @ottomata and we decided to abandon this approach and just do `conda install conda`.

PR #39, opened 2022-11-19
Update documentation and warnings

nshahquinn:


nshahquinn:
It's not that important to review the changes to the documentation (although it's welcome if you have time).

However, I've made some changes to the warnings and notices which are more important since they touch the code. My main objectives there were to increase clarity by adding spacing, removing unnecessary detail, and making informational notices less prominent in Jupyter notebooks (by using `print` instead of `print_err`). I also removed our use of `warnings.warn`; I like the theory of using it, but in practice it adds a lot of extra detail to warnings, which in our case isn't useful.

wmfdata/__init__.py
```diff
@@ -9,21 +11,33 @@
...
-        utils.print_err(update_message)
+        update_command = f"pip install --upgrade git+{metadata.source}.git@release"
+        if conda.is_anaconda_wmf_env():
+            update_command += " --ignore-installed"
+
+        utils.print_err(
+            f"You are using Wmfdata v{metadata.version}, but v{remote['version']} is available.",
+            f"To update, run `{update_command}`.",
+            f"To see the changes, refer to {metadata.source}/blob/release/CHANGELOG.md.",
+        )
```
xabriel: Could we modify `check_remote_version()` to also calculate whether the newer version has breaking changes, and include that as part of the notice? That way the potential impact on the user is more explicit, especially if they don't know about semantic versioning.

xabriel (approved):
Great stuff @nshahquinn!

+1, with one suggestion below.

wmfdata/__init__.py
nshahquinn: I actually had the same idea, but thought it might be over-engineering! Since you had the same idea, that suggests it's not. I just added it. 

For what it's worth, though, in the long term I think we should drop this homegrown upgrade system and rely on Conda's normal upgrade mechanics. It will mean not prompting users about new versions but I think a reasonable trade-off for just doing things the standard way.

xabriel (approved):
LGTM!

PR #40, opened 2022-11-22
Expand and update quickstart notebook

nshahquinn:


nshahquinn:
These changes only affect documentation, so I'm going to self-merge. I did make some changes based on good feedback from @bearloga on Slack.

PR #42, opened 2023-01-24
mariadb: allow setting sql query parameters

supertassu:


nshahquinn:
Hi @supertassu! Thank you for the PR.

Can you tell me why this is useful for you? If I understand correctly, this would allow parameterizing a single query, so you could accomplish the same thing even more simply by using f-strings. For example:
```python
import wmfdata as wmf

user_names = ["A", "B", "C"]

for user_name in user_names:
    wmf.mariadb.run(f"""
        SELECT *
        FROM user
        WHERE user_name = {user_name}
    """, "enwiki")
```

Now that I've written it, I realize it wouldn't work now that the actor table migration is complete. But hopefully it still makes the point 😊

supertassu:
Unless there's some magic going on that I don't see, using f-strings would open the code up for [SQL injection](https://en.wikipedia.org/wiki/SQL_injection) attacks when working with untrusted data (like I was when creating this PR).

nshahquinn:
Ah, that's true! F-strings don't protect against SQL injection.

However, this library is designed for data analysis. I can't think of any times when my work or other analysts' work involved passing untrusted data into queries. Even if such a situation came up, I don't think it could cause any damage since (1) we don't have write access to important data and (2) the output of our queries isn't automatically propagated anywhere or shared publicly without review. We do very occasionally write automated jobs that publish their output, but again I don't think that has ever involved using untrusted data as input.

Can you tell me more about the work you were doing with untrusted data? 😊

supertassu:
I was querying some data to estimate the impact of a bug in [CentralAuth](https://www.mediawiki.org/wiki/Extension:CentralAuth), and as a part of that I needed to query some user names from the CA metadata database and then perform some queries based on those usernames on individual wiki databases. (Not sharing the exact details on a public venue, but happy to explain more somewhere private.) In that case those usernames were untrusted data.

Additionally I'd argue that it's important to always try to escape parameters, not just when there's a possibility of exploitation. That way properly handling input data becomes a habit and you then don't accidentally forget to do that when it really matters.

nshahquinn:
> I was querying some data to estimate the impact of a bug in [CentralAuth](https://www.mediawiki.org/wiki/Extension:CentralAuth), and as a part of that I needed to query some user names from the CA metadata database and then perform some queries based on those usernames on individual wiki databases. (Not sharing the exact details on a public venue, but happy to explain more somewhere private.) In that case those usernames were untrusted data.

Thank you, this is plenty of information! If user names and page titles are untrusted, then yes, we do use untrusted data more than I was saying. I was thinking of truly free-form data like page content or search queries; I feel confident saying that is never passed into Wmfdata queries. 

> Additionally I'd argue that it's important to always try to escape parameters, not just when there's a possibility of exploitation. That way properly handling input data becomes a habit and you then don't accidentally forget to do that when it really matters.

I see what you're saying. However, I just don't think it's worth the cost. The vast majority of people who use this library never do "real" software development, and it doesn't seem reasonable to add a new and more difficult method of parameter binding and to train them all to use it just in case.

And, as I've said, when it comes to direct risk rather than the indirect risk via bad habits, I just don't think it exists.

So, I'm going to decline this. I realize this is small change, but it would still expand our public API and take us further away from having "just one way to do it".

If you want to discuss this more, please do. Just know that I will be out next week so my response will be delayed.

(For what it's worth, even though this library is the only "real" software development I've ever done, I would like to do things like, say, contributing to MediaWiki one day. So I do appreciate the contact with proper software development practices and  the practice at thinking through the risks of SQL injection)

PR #43, opened 2023-06-17
Use new method to set Pandas's display.max_colwidth

nshahquinn:
In Pandas 2.0, display.max_colwidth can no longer be disabled by setting it to `-1`. `None` must be used instead.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,10 @@ setup(
         # This is the latest version which supports MariaDB Connector/C 3.1.16,
         # which is the version currently available on the analytics clients
         "mariadb==1.0.11",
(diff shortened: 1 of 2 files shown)


xabriel (approved):
Shouldn't this PR be called `Bump pandas to >=1.0` ? The `display.max_colwidth` issue seems incidental to the bump.

nshahquinn:
> Shouldn't this PR be called `Bump pandas to >=1.0` ? The `display.max_colwidth` issue seems incidental to the bump.

Well, normally I would agree. But the bump is really only for form since Wmfdata is only designed to work in our hosted Jupyter and that already starts with 1.4.3.

I will note it in the commit description, though.

PR #45, opened 2023-09-01
Pin Urllib3 below 2.0 to avoid Presto errors

nshahquinn:
Urllib3 2.0 removes support for finding the host name in the key certificate's commonName field during TLS authentication, which makes it impossible for presto-python-client to connect to our Presto coordinator.

Bug: T345309

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,12 @@ setup(
         "requests_kerberos",
         "presto-python-client",
         "pyhive[hive]",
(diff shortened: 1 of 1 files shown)


xabriel:
I also hit this recently in https://phabricator.wikimedia.org/T340861#9100284.

LGTM!

nshahquinn:
@xabriel thank you for taking a look at it, and so quickly too! 😁

PR #47, opened 2023-11-28
Update the ca_bundle used for presto connections

tullis:
Until now we have been using Puppet certificates to encrypt connections to Presto. We will shortly be using certificates that are generated by the PKI system instead.

As such, wmfdata-python needs to know about our root CA as well as the puppet CA. This change enables that by using a new, combined CA bundle.

Once this change in in place, we can safely update the presto certificates and wmfdata-python will continue to work.

Bug: T345482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Next
+* The CA bundle that is used for establishing a TLS connection with presto has been updated to the new combined bundle. This supports certificates signed by the
+legacy Puppet 5 built-in certificate authority, as well as the newer certificates signed by the WMF PKI system.
+
 # 2.0.1 (18 September 2023)
 * Urllib3 is now pinned below 2.0 to avoid errors when querying Presto ([T345309](https://phabricator.wikimedia.org/T345309)).
 * Matplotlib is no longer specified as a dependency, since the `charting` module was removed in 2.0.
--- a/wmfdata/presto.py
+++ b/wmfdata/presto.py
@@ -36,7 +36,7 @@ def run(commands, catalog="analytics_hive"):
         config="/etc/krb5.conf",
         service_name="presto",
         principal=f"{USER_NAME}@WIKIMEDIA",
-        ca_bundle="/etc/ssl/certs/Puppet_Internal_CA.pem"
+        ca_bundle="/etc/ssl/certs/wmf-ca-certificates.crt"
     )
 
     connection = prestodb.dbapi.connect(


xabriel:
> We will shortly be using certificates that are generated by the PKI system instead.

@tullis This is a breaking change, correct?

tullis:
> > We will shortly be using certificates that are generated by the PKI system instead.
> 
> @tullis This is a breaking change, correct?

No, this is safe to deploy at any time.

The new bundle contains both the old Puppet CA certificate and the new PKI root certificate, so wmfdata-python will support both certificate authorities after this change.

xabriel:
> > > We will shortly be using certificates that are generated by the PKI system instead.
> > 
> > 
> > @tullis This is a breaking change, correct?
> 
> No, this is safe to deploy at any time.
> 
> The new bundle contains both the old Puppet CA certificate and the new PKI root certificate, so wmfdata-python will support both certificate authorities after this change.

And `an-coord1001.eqiad.wmnet` will continue working for older `wmfdata` versions, even when you push https://gerrit.wikimedia.org/r/709713 to production?

tullis:
> > > > We will shortly be using certificates that are generated by the PKI system instead.
> > > 
> > > 
> > > @tullis This is a breaking change, correct?
> > 
> > 
> > No, this is safe to deploy at any time.
> > The new bundle contains both the old Puppet CA certificate and the new PKI root certificate, so wmfdata-python will support both certificate authorities after this change.
> 
> And `an-coord1001.eqiad.wmnet` will continue working for older `wmfdata` versions, even when you push https://gerrit.wikimedia.org/r/709713 to production?

Good question. It's actually dependent on the puppet 7 upgrade, which is itself linked to the bullseye upgrade.
We discovered yesterday that the older wmfdata-python is already broken on stat1009, which has been upgraded to bullseye and puppet 7.

[Slack thread](https://wikimedia.slack.com/archives/CLKDS4MG9/p1701196404219089)

Currently, the other 5 stat servers are still on Debian buster and puppet 5.
There is a plan to upgrade them, and there is a plan to refresh an-coord1001 with an-coord1003, so there are lots of moving parts.

Effectively, the old versions of wmfdata-python will keep working until either a stat host gets upgraded to bullseye, or presto moves to a new rust running bullseye.

I think that, given wmfdata-python is already broken on stat1009, we should probably try to get this change deployed pretty soon.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.0.2 (28 November 2023)
```
xabriel: The release process is separate from specific contributions.

So perhaps we can rename this to `# Next` for now?

wmfdata/metadata.py
```diff
@@ -4,4 +4,4 @@
 # installed.
 
 source = "https://github.com/wikimedia/wmfdata-python"
-version = "2.0.1"
+version = "2.0.2"
```
xabriel: Let's remove this change for now, and later we can do a release via https://github.com/wikimedia/wmfdata-python/blob/main/docs/developing.md#releasing-a-new-version

tullis: Done.

CHANGELOG.md
tullis: Done.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Next (28 November 2023)
```
xabriel: Let's remove the date as well as that will be done when we cut release.

xabriel (approved):
One minor comment above.

Thanks for the fix @tullis !

CHANGELOG.md
tullis: Done. Thanks.

PR #46, opened 2023-10-16
T341589: Improve formatting of df_to_remarkup / add kwargs

andrewtavis-wmde:
## Changes

- `df_to_remarkup` now makes use of `pandas.DataFrame.to_markdown` to dramatically simplify the code
  - Adding `.replace(":", "-")` to the result is the only required change for Phabricator formatting
- This change also allows for kwargs for `pandas.DataFrame.to_markdown` and the Python package `tabulate` to be passed for formatting the results (including row numbers and formatting float columns)
- The doc string for the function has been changed to reflect the new usage and includes examples for the `index` and `floatfmt` kwargs
- Minor changes to indentation in `utils.py` in places where two spaces were being used instead of four

## Notes

- Please see [T341589](https://phabricator.wikimedia.org/T341589) for the discussion :)

PR #48, opened 2023-12-05
Fix tests for df_to_remarkup().

xabriel:
Fixes broken tests after #46.


andrewtavis-wmde:
Thanks for the fix, @xabriel! Sorry it needed to be done.

PR #49, opened 2023-12-15
Retry presto.run with the alternative hostname

tullis:
We are preparing to switch the kerberos principal used for the presto coordinator from an-coord1001.eqiad.wmnet to
analytics-presto.eqiad.wmnet

The DNS hostname will also change and will be a CNAME alias for one of several coorinator candidates.

In order to avoid a breaking change in wmfdata-python, we will need to be able to support both of these hiostnames and principals for a short time. In order to do this we are adding a try/except mechanism to the presto.run method. If it fails to connect to the old name, it will try the new name.

The old name will be removed in due course, once the switch has been made.

Bug: T345482

### DIFF
--- a/wmfdata/presto.py
+++ b/wmfdata/presto.py
@@ -1,6 +1,7 @@
 import os
 import warnings
 
+import dns.resolver
 import pandas as pd
 import prestodb
 from requests.packages import urllib3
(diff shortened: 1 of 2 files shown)


wmfdata/presto.py
```diff
@@ -39,15 +39,26 @@ def run(commands, catalog="analytics_hive"):
...
-    connection = prestodb.dbapi.connect(
-        catalog=catalog,
-        host="an-coord1001.eqiad.wmnet",
-        port=8281,
-        http_scheme="https",
-        user=USER_NAME,
-        auth=PRESTO_AUTH,
-        source=f"{USER_NAME}, wmfdata-python"
-    )
+    try:
```
xabriel: Can we add a comment here to explain the temp code?

tullis:
I am having a little rethink of the approach here, given what I read in:
http://prestodb.io/blog/2022/04/15/disggregated-coordinator/

We have so far only ever run a single presto coordinator on an-coord1001, so my initial plan had been to use a big-bang approach to making the switch. However, now that I read about disaggregated coordinators, perhaps the better approach is to start presto coordinators on an-coord100[3-4] and have them added to the cluster of coordinators.

I'll convert this to a draft whilst I consider our options.

tullis:
This wouldn't have worked, due to the way that kerberos was (and still is) configured.

PR #50, opened 2024-01-26
Allow wmfdata to use a DNS CNAME for the presto coordinator

tullis:
We wish to use a CNAME for analytics-presto.eqiad.wmnet in order to make
it possible to migrate the presto coordinator role from one host to
another.

### DIFF
--- a/wmfdata/presto.py
+++ b/wmfdata/presto.py
@@ -13,12 +13,6 @@ from wmfdata.utils import (
 
 PRESTO_HOST = "analytics-presto.eqiad.wmnet"
 
-# Disable a warning issued by urllib3 because the certificate for an-coord1001.eqiad.wmnet
-# does not specify the hostname in the subjectAltName field (T158757)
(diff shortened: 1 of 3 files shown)


wmfdata/presto.py
```diff
@@ -1,6 +1,7 @@
 import os
 import warnings
 
+import dns.resolver
```
brouberol, line 4: Note: the `dnspython` package needs to be added to the `wmfdata` dependencies for this to work, in the `install_requires` in [`setup.py`](https://github.com/wikimedia/wmfdata-python/blob/main/setup.py#L20)

xabriel:
Looks good to me after fix suggested by @brouberol.

brouberol (approved):

xabriel (approved):
LGTM!

nshahquinn (approved):
Looks good to me! I tested Presto querying and ran the test suite, and everything worked.

PR #51, opened 2024-01-30
Unpin Urllib3 version

nshahquinn:
Now that the Presto infrastructure has been upgraded to use TLS certificates with the subjectAltName filled out, we no longer need to pin Urllib3 below v2.0.

Bug: T345483

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,12 +34,7 @@ setup(
         "requests_kerberos",
         "presto-python-client",
         "pyhive[hive]",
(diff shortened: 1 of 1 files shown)


nshahquinn:
I've tested with Urllib3 v2.2.0 and the updated Presto code works fine.

xabriel (approved):
+1 and merge to unblock @nshahquinn.

nshahquinn:
Thanks @xabriel!

PR #52, opened 2024-02-18
Remove code handling now-removed anaconda-wmf environments

nshahquinn:
Bug: [T346707](https://phabricator.wikimedia.org/T346707)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# Next
+* `spark.get_active_session` has been deprecated, since we are now using Spark 3, which has an equivalent built-in function, `pyspark.sql.SparkSession.getActiveSession`.
+
 # 2.3.0 (30 January 2024)
 * The `presto` module now follows the DNS alias `analytics-presto.eqiad.wmnet` to connect to the current Presto coordinator, rather than being hardcoded to connect to a particular coordinator. This allows Wmfdata to automatically adapt when the coordinator role is switched to a new server ([T345482](https://phabricator.wikimedia.org/T345482)).
 * The version pin and warning handling code for Urllib3 has been removed, thanks to the updated certificate bundle added in 2.2.0.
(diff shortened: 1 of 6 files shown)


nshahquinn:
I don't consider this a blocker for merging but this code creates one oddity which bugs me. Please let me if you have any insights!

This commit causes a new warning message (almost invisible in the torrent of logspam) to appear in the output when you run `wmfdata_tests/tests.py`:
```
...
TEST PASSED: Spark Session APIs
usage: tests.py [-h] hive_database
tests.py: error: unrecognized arguments: --all --json
...
[Spark session startup logspam]
...
TEST PASSED: Read via Spark      
```

This makes it seem like something is going wrong inside `test_read_via_spark` with a call to `conda.info()`, which includes `condacli("info", ["--all", "--json"]`. But I have no idea how `conda.info` might get called or why it would cause that warning message even if it did (since the function works fine on its own).

I could not replicate this by running the code from `test_read_via_spark` both in a notebook and in a Python script, which underscores that this is not an actual problem.

But, still, super weird, and I'd feel great if anyone else managed to figure it out!

xabriel (approved):
LGTM!

xabriel:
> This commit causes a new warning message (almost invisible in the torrent of logspam) to appear in the output when you run `wmfdata_tests/tests.py`:

So you tried running the tests without the changes from this MR and you could not repro?

> This makes it seem like something is going wrong inside `test_read_via_spark` with a call to `conda.info()`, which includes `condacli("info", ["--all", "--json"]`. But I have no idea how `conda.info` might get called or why it would cause that warning message even if it did (since the function works fine on its own).

We don't pin `conda` on this repo, but we do pin it downstream in `conda_analytics`, so perhaps the API changed.

nshahquinn:
@xabriel thank you for the speedy review!

> So you tried running the tests without the changes from this MR and you could not repro?

Yes, exactly. If I switch back to the previous commit, leaving everything else the same, and run the tests again, I don't get the message.

PR #53, opened 2024-10-26
Miscellaneous minor improvements

nshahquinn:
See the commit messages and associated Phab tasks for details.

### DIFF
--- a/wmfdata/spark.py
+++ b/wmfdata/spark.py
@@ -65,15 +65,8 @@ ENV_VARS_TO_PROPAGATE = [
 
 def get_active_session():
     """
-    DEPRECATED: Returns the active session if there is one and None otherwise.
-
-    This is a holdover from before Spark 3, when there was no native getActiveSession function.
+    Returns the active session if there is one and None otherwise.
     """
-    print_err(
-        "get_active_session has been deprecated and will be removed in the next major version. "
-        "Use pyspark.sql.SparkSession.getActiveSession instead."
-    )
-
     return SparkSession.getActiveSession()
 
 
@@ -112,7 +105,7 @@ def create_custom_session(
     """
     check_kerberos_auth()
 
-    session = SparkSession.getActiveSession()
+    session = get_active_session()
     if session:
         session.stop()
 
@@ -252,7 +245,7 @@ def run(commands: Union[str, List[str]]) -> pd.DataFrame:
 
     commands = ensure_list(commands)
 
-    session = SparkSession.getActiveSession()
+    session = get_active_session()
     if not session:
         session = create_session() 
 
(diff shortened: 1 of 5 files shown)


nshahquinn:
I included them all together because they're so small and separate merge requests would have encountered conflicts in the changelog (there's probably a better way of dealing with that issue, but I'll leave that for another day).

Please rebase instead of squashing since these commits are logically separate.

wmfdata/spark.py
```diff
@@ -65,15 +65,8 @@
 
 def get_active_session():
     """
-    DEPRECATED: Returns the active session if there is one and None otherwise.
```
xabriel, line 68: Why do we need a wrapper function when `SparkSession.getActiveSession()` is available and [part of PySpark API](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.SparkSession.getActiveSession.html#pyspark.sql.SparkSession.getActiveSession)?

The deprecation I think was the right call.

xabriel:
> I included them all together because they're so small and separate merge requests would have encountered conflicts in the changelog (there's probably a better way of dealing with that issue, but I'll leave that for another day).

I understand the convenience of a single MR, but as you mention later:
> Please rebase instead of squashing since these commits are logically separate.

So you are shifting the burden of maintenance to the committer/reviewer. Fine for today, but let's make separate MRs for logically separate contributions in the future.

nshahquinn:
@xabriel thank you for the review!
> I understand the convenience of a single MR, but as you mention later:
> 
> > Please rebase instead of squashing since these commits are logically separate.
> 
> So you are shifting the burden of maintenance to the committer/reviewer. 

Well, "rebase instead of squash" just refers to which merge option to pick in case you decide to press big green merge button, so it's not any extra _work_ for the reviewer, but I suppose having to communicate which option to pick is unnecessary overhead.

> Fine for today, but let's make separate MRs for logically separate contributions in the future.

Fair, it's good to always do it the same way even for small things. I'll stick to separate MRs in the future.

wmfdata/spark.py
nshahquinn: In my experience, you can do most analysis with Pyspark just using the methods attached to the `SparkSession` object, without ever importing `pyspark` directly (for example, I never imported it directly in [this big notebook](https://gitlab.wikimedia.org/nshahquinn-wmf/editor-month-history-shift/-/blob/main/analysis.ipynb)).

However, if you start the session by using `wmf.spark.run`, you aren't returned that session object. In that case, if you want to start use the PySpark APIs with the existing session, you would have to do: `import pyspark; spark = pyspark.sql.SparkSession.getActiveSession()`. Obviously, not a big deal to do, but I do always find myself having to look up the exact command, whereas I find `spark = wmf.spark.get_active_session()` very ergonomic and obvious because it's parallel to `spark = wmf.spark.create_session()`. 

I'm generally against wrappers for simple operations; that's why I deprecated it in the first place. But after I did that, there were several instances where I found myself using and appreciating it, so now I feel like this particular wrapper is useful.

xabriel: Makes sense.

PR #55, opened 2024-10-31
Suppress Pandas UserWarning with decorator

nshahquinn:
Bug: T324135

### DIFF
--- a/wmfdata/mariadb.py
+++ b/wmfdata/mariadb.py
@@ -2,13 +2,12 @@ import atexit
 import getpass
 import grp
 import subprocess
-import warnings
 
 import mariadb
 from mariadb.constants import FIELD_TYPE
 import pandas as pd
 
-from wmfdata.utils import ensure_list
+from wmfdata.utils import ensure_list, suppress_pandas_dbapi_warning
 
 
 # Set up converter for binary data
@@ -90,6 +89,7 @@ def connect(db, use_x1=False):
     return connection
 
 # To-do: provide an easy way to get lists of wikis
+@suppress_pandas_dbapi_warning
 def run(
   commands, dbs, use_x1=False, format="pandas", date_col=None,
   index_col=None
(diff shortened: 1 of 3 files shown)


wmfdata/utils.py
```diff
@@ -41,6 +42,19 @@ def pd_display_all(df):
     ):
         display(df)
 
+def suppress_t324135(fn):
```
nshahquinn: I would prefer a name that makes sense even if you don't look at the task or the docstring. How about `suppress_pandas_dbapi_warning`?

wmfdata/mariadb.py
```diff
@@ -139,19 +139,12 @@ def run(
         # then turn them into a data frame; this would let us avoid using
         # Pandas's complex SQL machinery.
         for command in commands:
-            try:
```
nshahquinn, line 142: Looks like you accidentally dropped this `try`.

nshahquinn (commented):
I found a couple things for you to revise, but otherwise it looks good! Thank you for taking the time to polish it further :grin:

If you can run the test script (`wmfdata-python/wmfdata_tests/tests.py {{your hive db}}`). If it finishes without any assertion errors, that's a pass.

wmfdata/utils.py
jsnshrmn: Makes sense!

wmfdata/mariadb.py
jsnshrmn: oopsydoodles; that's what I get for coding when I can't execute. That's addressed now that I have the proxy configured correctly on the stat host. Thanks!

jsnshrmn:
> I found a couple things for you to revise, but otherwise it looks good! Thank you for taking the time to polish it further 😁
> 
> If you can run the test script (`wmfdata-python/wmfdata_tests/tests.py {{your hive db}}`). If it finishes without any assertion errors, that's a pass.

I can't connect to hive yet (I just made a kerberos account request), but I've made the suggested changes.

nshahquinn:
> I can't connect to hive yet (I just made a kerberos account request), but I've made the suggested changes.

Ah, makes sense! I just ran it and everything looks good 🎉

